PR #6, opened 2026-06-14
feat(es-indexer): Kafka consumer + OpenSearch bulk writer + index mapping

yujiawei:
## Summary

Phase 4 of the message-search pipeline: implement the standalone `es-indexer`
service end-to-end — Kafka `octo.message.v1` → OpenSearch — in this new
one-repo-one-image repo. Builds on the merged scaffold (`main`).

The binary idles with no backend connection unless `ES_INDEXER_ENABLED=true` and
brokers + ES are configured, so deploying it is a no-op until explicitly enabled.

## What's in this PR

**Reusable ES writer (`internal/esindex`)**
- `opensearch-go/v3` `_bulk` upsert, doc `_id = message_id` (idempotent sink —
  duplicate delivery overwrites the same doc → effectively-once over
  at-least-once). Per-item HTTP status mapped back to `BulkItemResult`
  (`Permanent()` = 4xx except 429); batch-level / missing-item failures are
  transient, never silently dropped.
- `EnsureIndex` bootstraps the index idempotently from an **embedded canonical
  mapping** (`mapping/octo-message.json`) with a `cjk_bigram` Chinese analyzer.
  That file is the single source the octo-deployment coordinated change
  references (override `content.analyzer` there for IK/smartcn if the managed
  cluster provides it). Decoupled from the consumer so the phase-6 backfill job
  imports `internal/esindex` directly.

**Kafka consumer (`internal/consumer`), satisfying C4**
- `FetchMessage` + `CommitInterval=0` (manual commit only; no `ReadMessage`).
- `schema_version` validated; unknown versions → DLQ, never silently consumed.
- Offset advances only to the **per-partition contiguous success prefix**; the
  first transient stops that partition's prefix, so a monotonic high-watermark
  commit can never confirm an unprocessed message. Partitions advance
  independently.
- Transient (429/5xx/network/batch-level) → **in-place retry of the same batch**
  (re-bulk only the still-unresolved subset, no new fetch) until resolved or ctx
  cancels. Permanent (4xx non-429 + unknown schema) → DLQ → offset crosses.
- **DLQ terminal escape**: bounded retry, then spill-to-disk + alarm + advance
  (when `INDEXER_DLQ_SPILL_DIR` is set) or a **fatal hard-stop** that halts the
  worker and pages (when not). A DLQ outage can never wedge the prefix forever,
  and hard-stop never silently crosses the poison pill.

Env-configured service wiring (`internal/consumer/service.go`, `cmd/es-indexer`).

## Tests (`-race`, golangci-lint clean)

Writer: all-success / mixed 4xx-429-5xx / batch failure / idempotent `_id` /
missing response item; `EnsureIndex` create-when-missing vs no-overwrite;
embedded-mapping validity. Consumer: unknown schema → DLQ; 4xx → DLQ → offset
crosses; **in-place transient retry then heal + commit**; **persistent transient
never commits past it**; 429 transient; DLQ retry-then-succeed; DLQ spill escape;
DLQ fatal hard-stop; per-partition prefix matrix (single/head-transient/
all-resolved/**multi-partition independent**/out-of-order); decision + DLQ unit
matrices.

## Cross-repo coordination

The ES index mapping/analyzer lives at `internal/esindex/mapping/octo-message.json`
and is embedded here; the **octo-deployment** change must provision the managed
OpenSearch template at the same contract version (`searchmsg.SchemaVersion`).
`/metrics` + lag/backlog instrumentation land in phase 7.


Jerry-Xin (approved):
The PR is in scope for `octo-search-indexer` and implements the expected Kafka-to-OpenSearch indexing path with solid coverage around retry, DLQ, and prefix commit behavior.

**💬 Non-blocking**
- 🟡 Warning: [internal/esindex/ensure_index.go](/Users/mlamp/.openclaw/tmp/pr-review-onbetx2k/repo/internal/esindex/ensure_index.go:34) treats `resp == nil` from the index existence check as “missing” and proceeds to create the index. Transport errors should probably fail fast instead of falling through to create.
- 🟡 Warning: [internal/consumer/service.go](/Users/mlamp/.openclaw/tmp/pr-review-onbetx2k/repo/internal/consumer/service.go:58) can leak the Kafka reader if `newKafkaDLQSink` fails after `newKafkaSource` succeeds. Close already-created resources on constructor errors.
- 🔵 Suggestion: [internal/consumer/consumer.go](/Users/mlamp/.openclaw/tmp/pr-review-onbetx2k/repo/internal/consumer/consumer.go:143) and [internal/consumer/dlq.go](/Users/mlamp/.openclaw/tmp/pr-review-onbetx2k/repo/internal/consumer/dlq.go:92) use plain sleeps, so cancellation waits for the backoff duration. A context-aware timer would improve shutdown behavior.
- 🔵 Suggestion: [internal/consumer/kafka.go](/Users/mlamp/.openclaw/tmp/pr-review-onbetx2k/repo/internal/consumer/kafka.go:98) enables DLQ topic auto-creation. That is convenient, but in production a typo can silently create the wrong topic.

**✅ Highlights**
- Manual commit and per-partition prefix logic are well isolated and tested.
- DLQ hard-stop versus spill behavior is explicit and covered.
- `searchmsg.Message` JSON fields match the strict OpenSearch mapping.
- Verified locally: `go test ./...`, `go test -race ./...`, and `go vet ./...` all pass.

mochashanyao (changes requested):
[Octo-Q · automated review]

Verdict: Request changes — blocking findings below (data-flow traced).

---

# octo-search-indexer PR#6 Review Report
**PR:** Mininglamp-OSS/octo-search-indexer#6
**Head SHA:** `71cdd21627bd8fad4110b764b017cda5c570854a`
**Scope:** feat(es-indexer): Kafka consumer + OpenSearch bulk writer + index mapping

---

## 1. Verification Summary

| Area | Status | Evidence |
|------|--------|----------|
| Contract field alignment (searchmsg.Message ↔ ES mapping) | ✅ | All 11 fields match by name, type, and JSON tag. `dynamic: strict` rejects unknown fields. |
| Kafka manual commit (C4: FetchMessage + CommitInterval=0) | ✅ | `kafka.go:52` sets `CommitInterval: 0`; `Fetch()` uses `FetchMessage`; `Commit()` uses `CommitMessages`. |
| Per-partition contiguous prefix commit | ✅ | `decision.go:62-92` sorts by offset per partition, breaks at first transient. Multi-partition test passes. |
| Transient in-place retry (no new fetch) | ✅ | `consumer.go:131-155` retry loop re-runs `resolvePass` on same batch; `p.sleep(p.transientBackoff)` between attempts. |
| Permanent → DLQ → offset crosses | ✅ | `consumer.go:196-210` classifies 4xx as permanent, routes to DLQ, marks `dispDLQResolved`. |
| DLQ terminal escape (spill / hard-stop) | ✅ | `dlq.go:87-121` bounded retry → spill-to-disk (if SpillDir) or `errDLQHardStop`. |
| Idempotent sink (_id = message_id) | ✅ | `writer.go:131` sets `action.Index.ID = msgs[i].MessageID`; uses `index` action (upsert). |
| Schema version validation | ✅ | `consumer.go:258-265` rejects `msg.SchemaVersion != searchmsg.SchemaVersion` → DLQ. |
| Bulk result alignment (request ↔ response) | ✅ | `writer.go:175-210` maps by index position; defensive check for missing items (Status=0, transient). |
| EnsureIndex idempotent (single instance) | ✅ | `ensure_index.go:34-50` checks Exists first, creates only on 404. |
| EnsureIndex idempotent (concurrent instances) | ❌ | Race condition: concurrent Create calls fail with `resource_already_exists_exception`. See P1 below. |
| Test coverage | ✅ | 28+ tests covering all C4 paths, multi-partition, DLQ escape, transient retry, mixed statuses. |

## 2. Findings

### P1 — EnsureIndex concurrent startup race causes service crash

**File:** `internal/esindex/ensure_index.go:34-50`
**Severity:** P1 (blocking)
**Diff-scope:** **new** — this file is entirely new in this PR.

**Problem:** `EnsureIndex` does a check-then-act: `Indices.Exists` → if 404 → `Indices.Create`. When 2+ service replicas start concurrently (normal Kubernetes deployment with `replicas >= 2`, rolling update, or pod reschedule), both see 404 and both attempt `Create`. The loser gets HTTP 400 `resource_already_exists_exception` from OpenSearch, which `opensearch-go/v3` surfaces as a non-nil error. This propagates through `Service.Run` → `run()` → `main()` → `os.Exit(1)`.

**Data-flow trace:**
1. `service.go:99` calls `s.writer.EnsureIndex(ctx)` at startup
2. `ensure_index.go:38` — `Indices.Exists` returns 404 for both instances
3. `ensure_index.go:46` — Instance A's `Indices.Create` succeeds (200)
4. `ensure_index.go:46` — Instance B's `Indices.Create` returns 400 + error
5. `ensure_index.go:48` — returns `fmt.Errorf("esindex: create index %q: %w", ...)`
6. `service.go:100` — returns wrapped error
7. `main.go:40` — `run()` returns error → `os.Exit(1)`

**Impact:** In any multi-replica deployment, one or more replicas crash-loop on startup. The index exists and is healthy — only the startup race kills the process. This makes the "deploy the indexer" path unreliable in production.

**Suggested fix:** After `Indices.Create` fails, re-check `Indices.Exists`. If the index now exists, treat as success:

```go
if _, cerr := client.Indices.Create(ctx, opensearchapi.IndicesCreateReq{...}); cerr != nil {
    // Re-check: another instance may have created it between our Exists and Create.
    if rresp, rerr := client.Indices.Exists(ctx, opensearchapi.IndicesExistsReq{Indices: []string{index}}); rerr == nil && rresp != nil && rresp.StatusCode == http.StatusOK {
        return nil // created by peer, idempotent success
    }
    return fmt.Errorf("esindex: create index %q: %w", index, cerr)
}
```

Alternatively, inspect the error for `resource_already_exists_exception`.

**R2 check:** New code in this PR. Not pre-existing. Not amplified — entirely introduced here.
**R1 check:** Makes a working path (multi-replica deployment) unavailable in production → P1.

---

### P2 — No jitter or exponential backoff in transient retry

**File:** `internal/consumer/consumer.go:140`
**Severity:** P2 (maintainability / operational risk)
**Diff-scope:** new

`p.sleep(p.transientBackoff)` uses a fixed 1-second delay (default) between in-place retries. Under sustained OpenSearch overload (429/503), all consumer instances retry at the same cadence → thundering herd. Consider adding jitter (`rand.Duration`) and/or exponential backoff with cap.

### P2 — DLQ sink auto-creates topics

**File:** `internal/consumer/kafka.go:97`
**Severity:** P2 (operational concern)
**Diff-scope:** new

`AllowAutoTopicCreation: true` on the DLQ producer means a typo in `KAFKA_DLQ_TOPIC` silently creates a wrong topic. In production, DLQ topics should be pre-provisioned with correct retention/partitioning. Consider making this configurable or defaulting to `false`.

### P2 — Sleep in retry loop is not context-aware

**File:** `internal/consumer/consumer.go:140`, `dlq.go:100`
**Severity:** P2 (graceful shutdown latency)
**Diff-scope:** new

Both `p.sleep(p.transientBackoff)` and `h.sleep(h.cfg.RetryBackoff * ...)` use `time.Sleep`, which cannot be interrupted by context cancellation. During graceful shutdown (SIGTERM), the process may wait up to `transientBackoff` (1s default) or `RetryBackoff * attempt` before noticing the cancellation. Consider using a `select` with `ctx.Done()` or a context-aware timer.

### P2 — Missing test for EnsureIndex concurrent create race

**File:** `internal/esindex/writer_test.go`
**Severity:** P2 (test gap)
**Diff-scope:** new

`TestEnsureIndex_CreatesWhenMissing` and `TestEnsureIndex_NoCreateWhenExists` cover the happy paths but not the race condition where Exists returns 404 and Create returns 400 (resource_already_exists). A test mocking this sequence would lock the fix for P1 above.

---

## 3. Suggestions

1. **EnsureIndex race (P1):** Add the re-check pattern described above. Also add a test that mocks Exists→404 then Create→400 then Exists→200 to lock the behavior.
2. **Transient backoff:** Add `math/rand` jitter: `backoff := p.transientBackoff + time.Duration(rand.Int63n(int64(p.transientBackoff)))`. Consider exponential cap at 30s.
3. **DLQ auto-topic:** Default `AllowAutoTopicCreation` to `false`; expose via env var `INDEXER_DLQ_AUTO_CREATE_TOPIC`.
4. **Context-aware sleep:** Replace `time.Sleep(d)` with `select { case <-time.After(d): case <-ctx.Done(): }` pattern.

## 4. Additional Observations

- **Contract alignment is solid.** All 11 `searchmsg.Message` fields map correctly to the ES mapping. `dynamic: strict` prevents silent field drift. The `content` field (nullable `*string`) correctly maps to ES `text` with `cjk_bigram` analyzer.
- **C4 offset semantics are correctly implemented.** Per-partition contiguous prefix, transient in-place retry without new fetch, DLQ terminal escape — all verified through code trace and test coverage. The `partitionCommitPoints` function correctly handles multi-partition independence and out-of-order input.
- **Test quality is high.** Tests use dependency injection (fake source, fake writer, fake DLQ sink) and cover all C4 paths including edge cases (persistent transient, 429 at head, batch failure, DLQ spill, DLQ hard-stop, multi-partition).
- **`go.mod` dependencies are reasonable.** `opensearch-go/v3 v3.1.0` and `kafka-go v0.4.51` are established libraries. Indirect deps (`klauspost/compress`, `pierrec/lz4`) are pulled by kafka-go.

## 5. Data-Flow Trace

| Consumed Data | Upstream Source | Verified Flow |
|---|---|---|
| `batch[i].Value` (Kafka message bytes) | `kafkaSource.Fetch()` → `kafka.Reader.FetchMessage()` → `m.Value` | ✅ Flows to `decodeAndValidate()` → `json.Unmarshal` |
| `parsed[i]` (searchmsg.Message) | `decodeAndValidate()` return | ✅ Flows to `writer.Bulk()` via `toBulk` slice; only schema-valid, still-transient items included |
| `bulkRes[j]` (BulkItemResult) | `osWriter.Bulk()` → `mapBulkResults()` | ✅ Aligned by index to `toBulk`; `bulkIdx` maps back to batch position |
| `res.Status` (HTTP status code) | `mapBulkResults()` → `resp.Items[i]["index"].Status` | ✅ Flows to `classifyBulk()` → `isPermanentStatus()` for transient/permanent decision |
| `dispositions[i]` (item disposition) | Set by `resolvePass()` based on bulk result or DLQ outcome | ✅ Flows to `partitionCommitPoints()` for per-partition prefix calculation |
| `partitionCommitPoints` return | Computed from `batch` + `dispositions` | ✅ Each point passed to `src.Commit()` → `kafka.Reader.CommitMessages()` |
| `mappingJSON` (embedded mapping) | `//go:embed mapping/octo-message.json` | ✅ Flows to `Indices.Create` body; `IndexMappingJSON()` returns defensive copy |
| `dlqRecord` (DLQ payload) | `buildDLQRecord()` from `fetchedMessage` + `BulkItemResult` | ✅ Serialized to JSON → `dlqSink.WriteDLQ()` → `kafka.Writer.WriteMessages()` |
| `cfg` (ServiceConfig) | `loadConfig()` from env vars | ✅ Flows through `NewService()` to all component constructors |

## 6. Blind-Spot Checklist (R5)

- **C1 — Dual-path parity:** N/A. This is a new service with no symmetric add/remove or subscribe/unsubscribe paths. The Kafka consumer has a single forward-only flow (fetch → process → commit).
- **C2 — Control-flow ordering / nested reuse:** Clear. `resolvePass` is called in a retry loop but operates only on still-transient items (no double-processing). `encodeBulkBody` and `mapBulkResults` are called once per bulk attempt. No nested composite calls detected.
- **C3 — Authorization boundary ≠ capability boundary:** N/A. This is a backend service consuming from a configured Kafka topic and writing to a configured ES index. No user-facing endpoints or tool exposure. Authorization is handled at the infrastructure level (Kafka ACLs, ES credentials).

---

[Octo-Q] verdict: REQUEST_CHANGES

**Reason:** P1 finding — `EnsureIndex` has a TOCTOU race condition that causes service crash in multi-replica deployments (concurrent `Indices.Create` calls fail with `resource_already_exists_exception`). This is new code introduced in this PR, production-reachable in any standard Kubernetes deployment with replicas >= 2, and makes the service unavailable. Fix is straightforward (re-check Exists after Create failure).

yujiawei (commented):
# Code Review — PR #6 (octo-search-indexer)

Independent review at head SHA `71cdd21627bd8fad4110b764b017cda5c570854a` (merge-base `d7774ca`). Verified locally: `go build ./...`, `go vet ./...`, and `go test ./...` all pass.

Overall this is a well-structured phase-4 implementation. The C4 reliability semantics (manual commit, per-partition contiguous-prefix offset advance, transient in-place retry, permanent→DLQ, DLQ terminal escape) are implemented cleanly and have strong test coverage. The `searchmsg.Message` contract aligns field-for-field with the strict ES mapping. There is one blocking correctness issue in `EnsureIndex` and a few non-blocking items.

## Verification summary

| Area | Status | Evidence |
|------|--------|----------|
| Contract ↔ mapping alignment | ✅ | All 11 `searchmsg.Message` fields match by name/type/JSON tag; `dynamic: strict` rejects drift (`mapping/octo-message.json`) |
| Manual commit (FetchMessage + CommitInterval=0) | ✅ | `kafka.go:49,57,71` |
| Per-partition contiguous-prefix commit | ✅ | `decision.go:63-91`; breaks at first transient per partition |
| Transient in-place retry, no new fetch | ✅ | `consumer.go:139-164`; only still-transient subset re-bulked |
| Permanent (4xx≠429) + unknown schema → DLQ → cross | ✅ | `consumer.go:171-221`, `decision.go:31-42`, `writer.go:67-72` |
| DLQ terminal escape (spill / hard-stop) | ✅ | `dlq.go:79-131` |
| Idempotent sink (`_id = message_id`) | ✅ | `writer.go:157` (`index` action upsert) |
| Bulk request↔response alignment | ✅ | `writer.go:185-220`; missing-item defensively treated as transient |
| EnsureIndex — single instance | ✅ | `ensure_index.go:34-50` |
| EnsureIndex — concurrent instances | ❌ | TOCTOU create race → startup crash (P1 below) |

## Findings

### P1 (blocking) — `EnsureIndex` check-then-create race crashes replicas on startup

**File:** `internal/esindex/ensure_index.go:33-50`

`ensureIndex` does `Indices.Exists` → on 404 → `Indices.Create`. When two or more replicas start concurrently (any standard multi-replica Deployment, rolling update, or pod reschedule against an index that does not yet exist), both observe 404 and both issue `Indices.Create`. OpenSearch lets one win; the loser receives HTTP 400 `resource_already_exists_exception`.

I confirmed this is not swallowed: in `opensearch-go/v3 v3.1.0`, `(*Client).do` returns a non-nil error whenever `resp.IsError()` (status ≥ 300) — see `opensearchapi/opensearchapi.go:110-112`. So the losing `Indices.Create` returns an error.

Data-flow trace:
1. `service.go:93` → `s.writer.EnsureIndex(ctx)` at startup
2. `ensure_index.go:34` → `Indices.Exists` returns 404 for both replicas
3. `ensure_index.go:44` → winner `Create` 200; loser `Create` 400 + error
4. `ensure_index.go:48` → returns wrapped error
5. `service.go:94` → `Run` returns error → `main.go:41` → `os.Exit(1)`

**Impact:** in a normal `replicas >= 2` deployment where the index is not pre-provisioned, the losing replica(s) crash on startup (CrashLoopBackOff). It self-heals on restart once a peer has created the index (subsequent `Exists` returns 200), but this directly undermines the "deploy the indexer" path this PR enables, and it violates the function's own documented idempotency contract (`// EnsureIndex 幂等创建目标索引`).

**Suggested fix:** treat "already exists" as idempotent success — re-check `Exists` after a `Create` failure, or detect `resource_already_exists_exception`:

```go
if _, cerr := client.Indices.Create(ctx, opensearchapi.IndicesCreateReq{Index: index, Body: bytes.NewReader(mappingJSON)}); cerr != nil {
    if rresp, rerr := client.Indices.Exists(ctx, opensearchapi.IndicesExistsReq{Indices: []string{index}}); rerr == nil && rresp != nil && rresp.StatusCode == http.StatusOK {
        return nil // created by a peer between our Exists and Create — idempotent success
    }
    return fmt.Errorf("esindex: create index %q: %w", index, cerr)
}
```

Add a test mocking `Exists→404`, `Create→400`, `Exists→200` to lock the behavior.

### P2 (non-blocking) — `Indices.Exists` transport error falls through to Create

**File:** `internal/esindex/ensure_index.go:36-43`

When `Indices.Exists` returns a transport error with `resp == nil` (DNS/connection failure), control falls through to `Create` rather than failing fast. In practice `Create` will also fail and surface an error, so this is not silently wrong, but failing fast on a nil-resp transport error would give a clearer signal and avoid a spurious create attempt during a cluster blip.

### P2 (non-blocking) — backoff sleeps are not context-aware

**Files:** `internal/consumer/consumer.go:145`, `internal/consumer/dlq.go:93`

`p.sleep(...)` / `h.sleep(...)` wrap `time.Sleep`, which cannot be interrupted by `ctx`. On SIGTERM the worker may wait up to `transientBackoff` (1s default) or `RetryBackoff*attempt` before noticing cancellation. A `select { case <-time.After(d): case <-ctx.Done(): }` would tighten graceful shutdown. (Adding jitter to `transientBackoff` would also reduce thundering-herd retries.)

### P2 (non-blocking) — DLQ producer auto-creates topics

**File:** `internal/consumer/kafka.go:98`

`AllowAutoTopicCreation: true` means a typo in `KAFKA_DLQ_TOPIC` silently creates a mistyped topic instead of failing. DLQ topics are usually pre-provisioned with deliberate retention/partitioning; consider defaulting this to `false` and/or exposing it via env.

### Nit — possible reader leak on constructor error

**File:** `internal/consumer/service.go:58-70`

If `newKafkaDLQSink` fails after `newKafkaSource` succeeds (e.g. empty `DLQTopic`), the already-created Kafka reader is not closed. The process exits via `os.Exit(1)` immediately after, so impact is effectively nil, but closing already-created resources on constructor error is tidier.

## Strengths

- Manual-commit + per-partition prefix logic is cleanly isolated in a pure-decision file (`decision.go`) and exhaustively unit-tested (single/multi-partition, head-transient, out-of-order input).
- Transient vs permanent classification is correct: 429 is transient, other 4xx permanent, 5xx/network/batch-level transient; missing bulk-response items are defensively treated as transient (never silently dropped).
- DLQ hard-stop vs spill escape is explicit and covered, so a DLQ outage can never wedge the prefix forever.
- Idempotent `_id = message_id` upsert gives effectively-once on top of at-least-once.

## Verdict

Requesting changes for the single P1 (`EnsureIndex` concurrent-create race). The P2 items and the nit are non-blocking and can be addressed at the author's discretion. Once the P1 race is handled idempotently, this is good to merge.

lml2468:
## ✅ Re-review — octo-search-indexer PR #6 (905de63): EnsureIndex race fixed, verified — approving

Re-cloned at 905de63. The concurrent-create race (mochashanyao / yujiawei / Steve / me) is fixed, and more thoroughly than the minimum ask.

### 🔴→✅ EnsureIndex TOCTOU idempotency
On Create failure, ensureIndex now:
1. Authoritative already-exists check (isAlreadyExists): success only when real HTTP response is 400 AND structured opensearchapi.Error (or root cause) is resource_already_exists_exception. Does NOT trust a bare substring unless gated by 400 — avoids the trap where a 5xx/proxy body contains the token.
2. Fallback re-check (existsNow): for divergent concurrent error shapes, re-runs Exists, returns nil only if the index is actually present (200); otherwise the real error propagates.
3. Genuine 5xx with index still absent fails loudly — no masking.

### Tests — cover every edge
All 8 EnsureIndex cases pass, incl: TOCTOURaceIsIdempotent, RaceFallbackReCheck, HTTP500BodyClaims400Propagates (5xx body with token NOT mistaken for success), 5xxWithTokenStillAbsentPropagates, RealCreateErrorPropagates, ExistsShortCircuits, CreatesWhenMissing, NoCreateWhenExists.

### Carried (verified earlier)
Bulk per-item handling, multi-partition contiguous-prefix offset safety, transient/permanent/DLQ classification all sound. Prior 🟡s (service.New leak on partial-create, DLQ auto-create) remain non-blocking follow-ups.

Textbook idempotent TOCTOU close-out. LGTM.

lml2468 (approved):
Race fixed at 905de63: isAlreadyExists trusts only real HTTP 400 + structured resource_already_exists_exception (substring gated by status so a 5xx body with the token isn't a false positive); existsNow fallback confirms actual presence; real 5xx-with-index-absent propagates. 8 EnsureIndex tests pass incl false-positive + still-absent edges. Bulk per-item + multi-partition offset safety remain correct. LGTM.

mochashanyao (approved):
[Octo-Q · automated review]

---

# octo-search-indexer PR#6 Review Report
**Head SHA:** `905de630a90f25ca50867570fd5040d91603950d`
**PR:** Mininglamp-OSS/octo-search-indexer#6
**Scope:** 20 files, +2397/-44 — Kafka consumer + OpenSearch bulk writer + index mapping (Phase 4)

---

## 1. Verification Summary

| Area | Status | Evidence |
|------|--------|----------|
| C4 manual commit (FetchMessage + CommitInterval=0) | ✅ | `kafka.go:52` CommitInterval=0; `kafka.go:63-72` FetchMessage only; `kafka.go:74-79` CommitMessages explicit |
| Per-partition contiguous prefix commit | ✅ | `decision.go:62-92` partitionCommitPoints: per-partition sort by offset, break at first transient |
| Transient in-place retry (no new fetch) | ✅ | `consumer.go:131-168` processBatch retry loop re-runs resolvePass on same batch; `consumer.go:175-218` resolvePass only re-bulks dispTransient items |
| Permanent (4xx) → DLQ → offset crosses | ✅ | `decision.go:38-49` classifyBulk: 4xx→permanent; `consumer.go:199-204` permanent→routePoison→dispDLQResolved; `decision.go:78-80` DLQResolved counts in prefix |
| Unknown schema_version → DLQ | ✅ | `consumer.go:247-253` decodeAndValidate checks SchemaVersion; `consumer.go:183-189` schemaInvalid→routePoison |
| DLQ terminal escape (spill / hard-stop) | ✅ | `dlq.go:89-113` Send: bounded retry→escape; `dlq.go:117-134` escape: SpillDir→spill+nil, else→errDLQHardStop |
| EnsureIndex TOCTOU idempotent | ✅ | `ensure_index.go:51-75` Exists→Create→isAlreadyExists(400+type)→existsNow fallback; commit 2 fixes P1 from commit 1 |
| Idempotent sink (_id = message_id) | ✅ | `writer.go:120-121` action.Index.ID = msgs[i].MessageID; `writer.go:107-109` index action (not create) = upsert |
| Field alignment: searchmsg.Message ↔ ES mapping | ✅ | All 11 JSON fields in `searchmsg.Message` match `octo-message.json` mapping 1:1 (types compatible, dynamic:strict) |
| Exponential backoff + jitter + ctx-aware | ✅ | `backoff.go:17-31` expJitterBackoff capped at 30s; `backoff.go:37-47` sleepCtx returns on ctx.Done() |
| Test coverage | ✅ | 25+ test functions covering all C4 paths, TOCTOU race, DLQ escape, multi-partition independence, backoff bounds |

## 2. Findings

### P2-1: Response bodies from opensearchapi client not explicitly closed

**Files:** `ensure_index.go:53-56` (Exists), `ensure_index.go:63-66` (Create)

The `client.Indices.Exists()` and `client.Indices.Create()` calls return response objects whose `.Body` is not explicitly drained/closed. If the opensearch-go v3 typed API does not auto-close bodies internally, this leaks HTTP connections over the service lifetime (EnsureIndex is called once at startup, so the leak is bounded to 2-3 requests — low practical impact, but worth verifying against the v3 API contract).

**Suggestion:** Verify whether opensearch-go v3 `opensearchapi` auto-closes response bodies. If not, add `defer resp.Body.Close()` after nil-checking.

### P2-2: `mapBulkResults` relies on positional alignment without `_id` cross-check

**Diff-scope:** new
**File:** `writer.go:170-200`

The `_bulk` response `Items` are mapped to input messages by array index. While OpenSearch guarantees order preservation for `_bulk`, a defensive `_id` match check (warn-log on mismatch) would guard against proxy bugs or future client changes that could silently misalign results.

**Suggestion:** Add a debug/warn log when `item.ID != msgs[i].MessageID` (non-blocking, observability only).

### P2-3: No maximum retry cap on `processBatch` in-place transient loop

**Diff-scope:** new
**File:** `consumer.go:131-168`

The `for attempt := 0; ; attempt++` loop has no upper bound. If a single message triggers persistent 5xx from OpenSearch (e.g., a mapping conflict that returns 5xx instead of 4xx due to a proxy), the partition is blocked indefinitely with 30s max backoff between retries. The alert fires on each batch-level error, but there's no circuit-breaker to escalate to hard-stop after N failed attempts.

**Suggestion:** Consider a configurable `maxTransientAttempts` (default e.g. 100) after which the persistent-transient items are treated as permanent (DLQ or hard-stop). This is a defense-in-depth measure; the current design is defensible since 5xx is genuinely transient by definition.

### P2-4: `Service.Close()` captures only the first close error

**Diff-scope:** new
**File:** `service.go:104-115`

If multiple resources fail to close, only the first error is returned. Subsequent errors are silently dropped.

**Suggestion:** Use `errors.Join` (Go 1.20+) or accumulate errors. Low priority — close errors are rarely actionable.

## 3. Suggestions

1. **Observability:** Add structured logging for key state transitions (batch size, transient count per retry, DLQ record count). The current `logAlerter` covers alerts but not routine operational metrics. Phase 7 `/metrics` is noted in the PR description.
2. **Mapping shard count:** `number_of_shards: 1` in `octo-message.json` is fine for initial deployment but may need tuning for high-throughput clusters. Consider documenting the scaling path (reindex or index template override).
3. **`fetchBatch` 50ms凑批 timeout:** The hardcoded 50ms is reasonable but could be made configurable if latency tuning is needed in production.

## 4. Additional Observations

- **Commit 2 quality:** The second commit (`905de63`) is an exemplary fix for the TOCTOU race in `EnsureIndex`. The `isAlreadyExists` function correctly requires HTTP 400 at the transport layer (not trusting body `status` field), and the `existsNow` fallback covers non-standard error shapes. The regression tests (HTTP 500 body claiming 400, 5xx with token in text) are thorough.
- **DLQ record preserves raw bytes:** `dlqRecord.Value` stores the original Kafka message bytes, enabling replay/re-ingestion. The base64 encoding overhead in JSON serialization is acceptable for DLQ volumes.
- **kafka-go offset semantics:** `CommitMessages(ctx, Message{Offset: N})` in kafka-go sends `N+1` to the broker internally — the code correctly passes the last successfully processed message's offset.
- **Test quality:** Tests use dependency injection (fakeSource, fakeWriter, fakeDLQSink) with `healAfter` for transient-then-recover scenarios. The multi-partition independence test (`TestPartitionCommitPoints_MultiPartitionIndependent`) directly validates the core C4 invariant.

## 5. Data-Flow Trace

| Consumed Data | Upstream Source | Verified Flow |
|---|---|---|
| `fetchedMessage.Value` | `kafka.Reader.FetchMessage()` → `kafka.go:65-72` | ✅ `json.Unmarshal` in `decodeAndValidate` (`consumer.go:247`) |
| `parsed[i]` (searchmsg.Message) | `decodeAndValidate` output | ✅ Schema version checked; invalid→DLQ; valid→`toBulk` |
| `encodeBulkBody` input | `parsed[i]` filtered by `dispTransient && !schemaInvalid` | ✅ Only unresolved, schema-valid items reach bulk |
| `mapBulkResults` alignment | `resp.Items[i]["index"]` by positional index | ✅ OpenSearch guarantees order; defensive check suggested (P2-2) |
| `classifyBulk(res)` | `BulkItemResult.OK` + `BulkItemResult.Permanent()` (isPermanentStatus) | ✅ 429→transient, 4xx→permanent, 5xx/0→transient |
| `dlqRecord` fields | `fetchedMessage` + `BulkItemResult` + schema flag | ✅ Original bytes preserved; reason/status set correctly |
| `dlqHandler.Send` | `json.Marshal(dlqRecord)` → `sink.WriteDLQ(key, value)` | ✅ Retry with expJitterBackoff; escape to spill or hard-stop |
| `EnsureIndex` mapping | `//go:embed mapping/octo-message.json` → `mappingJSON` | ✅ Valid JSON, all 11 fields match searchmsg.Message |
| `partitionCommitPoints` | `batch[]` + `dispositions[]` grouped by `{topic, partition}` | ✅ Per-partition sort, break at first transient |
| `Service.Close()` resources | `source`, `dlqSink`, `writer` | ✅ All three closed; first error captured (P2-4) |

## 6. R5 Blind-Spot Checklist

- **C1 — Dual-path parity:** N/A. This is a unidirectional pipeline (Kafka→OpenSearch). No symmetric add/remove or subscribe/unsubscribe paths. The DLQ path is a one-way escape.
- **C2 — Control-flow ordering / nesting:** Clear. `expJitterBackoff` and `sleepCtx` are shared utilities used in both DLQ retry and transient retry without nesting or double-application. `resolvePass` correctly re-bulks only `dispTransient` items on retry (no double-processing of resolved items).
- **C3 — Authorization boundary ≠ capability boundary:** N/A. No user-facing endpoints. All credentials (Kafka SASL, OpenSearch basic auth) are env-configured. `INDEXER_DLQ_SPILL_DIR` is env-controlled with no external input. `AllowAutoTopicCreation=false` on DLQ producer prevents silent misconfiguration.

---

**[Octo-Q] verdict: APPROVE**

No P0/P1 findings. The C4 reliability semantics are correctly implemented with thorough data-flow integrity (contiguous prefix commit, in-place transient retry, DLQ terminal escape). The TOCTOU fix in commit 2 is well-engineered. Four P2 observations are non-blocking suggestions for hardening and observability. Test coverage is comprehensive and directly validates the core invariants.

Jerry-Xin (approved):
The PR is in scope for `octo-search-indexer`: it implements the Kafka-to-OpenSearch indexer functionality this repository exists to ship. I did not find blocking correctness, security, or data-loss issues.

🔴 Blocking

None.

💬 Non-blocking

🟡 Warning — A transient in one fetched batch stalls the whole reader, not only that partition. `Run` does not fetch another batch until `processBatch` returns, and `processBatch` retries the unresolved subset in place until all transients clear (`internal/consumer/consumer.go:81`, `internal/consumer/consumer.go:139`). Commits are computed per partition, which is good, but liveness is still worker-wide. If one partition has a long-lived transient, other assigned partitions will not receive newly fetched messages. Consider partition-scoped retry state or pausing only the affected partition in a later iteration.

🟡 Warning — Batch-level HTTP failures are always converted to `Status=0` transient retries (`internal/esindex/writer.go:144`). That is safe for data retention, but permanent batch-level failures such as oversized requests or unsupported media types can retry forever. The current alert helps, but it would be useful to classify known non-retryable batch HTTP statuses or at least expose stronger operational guidance.

🔵 Suggestion — `decodeAndValidate` treats malformed JSON and unknown `schema_version` identically, and `buildDLQRecord` records both as `unknown_schema_version` (`internal/consumer/consumer.go:251`, `internal/consumer/consumer.go:263`). Splitting those reasons would make DLQ triage and replay safer.

🔵 Suggestion — The bulk request does not explicitly set an NDJSON content type (`internal/esindex/writer.go:140`). The client currently defaults request bodies to `application/json`; many OpenSearch deployments accept that, but explicitly setting `application/x-ndjson` would better match the Bulk API contract and make proxies stricter about content types less surprising.

✅ Highlights

The core offset discipline is well covered: manual `FetchMessage`, `CommitInterval=0`, per-partition prefix calculation, and in-place retry avoid crossing unresolved transient messages.

The ES writer is cleanly separated from the consumer and has focused tests for mixed per-item statuses, idempotent `_id`, missing bulk response items, and index bootstrap races.

DLQ terminal escape is explicit and tested for retry-success, spill, and hard-stop paths.

Verification: `go test ./...` and `go test -race ./...` both pass locally.

lml2468:
## Follow-up on the retry-bounding 🟡s (Jerry-Xin + mochashanyao) — verified, refining my earlier praise

I approved at 905de63 (race fixed) and that stands. But Jerry-Xin's two retry 🟡s are real and sharpen a limitation my earlier review under-stated:

### Real: batch-level failure is unconditionally transient (writer.go:144-146)
w.client.Bulk(...) returning err (network/timeout/non-2xx whole response) -> batchFailure marks EVERY item transient (Status=0), and processBatch retries the batch forever (bounded only by ctx cancel). For a genuine transient (ES briefly down, 503) that's correct at-least-once behavior. But a PERMANENT batch-level failure — 401 after credential rotation, a 400 malformed request, a persistent misconfig — is also marked transient and loops indefinitely, never reaching DLQ. This is the accurate version of mochashanyao's "infinite loop" 🟡: not that retry is unbounded (correct for transients), but that batch-level errors aren't classified permanent-vs-transient the way per-item errors are (per-item already does 429->transient, other-4xx->permanent->DLQ). Apply the same split to the batch-level error (400/401/403 batch responses -> permanent -> DLQ/fail-fast).

### Real: cross-partition head-of-line blocking (Jerry-Xin)
My earlier praise of the per-partition contiguous-prefix commit is correct, but it only governs offset safety, not processing parallelism. processBatch retries one batch synchronously, and a batch spans multiple partitions — so one partition's transient item blocks the whole batch's in-place retry, stalling the other partitions in that batch. Offset-commit is per-partition-safe; throughput is not per-partition-isolated.

### Not a defect: the unbounded retry itself (mochashanyao #3 as stated)
For true transients, no-max-retries is the right at-least-once choice (a max would drop data or mis-DLQ a recoverable item). Exit is ctx cancel/SIGTERM, backoff is exponential+jittered+ctx-aware. So "infinite loop" as a blanket concern doesn't hold — the precise risk is only the batch-level-permanent case.

### Net
None block — race blocker fixed, per-item classification sound, these are robustness improvements under specific failure modes. Follow-ups: (1) classify batch-level errors permanent-vs-transient like per-item; (2) consider per-partition processing isolation if HOL blocking bites; (3) cheap hardening: response-body close (mochashanyao #1) + explicit NDJSON content-type (Jerry-Xin). APPROVE stands; tracked follow-ups, not merge gates.

yujiawei (commented):
# Code Review — PR #6 (octo-search-indexer)

Reviewed at head SHA `905de630a90f25ca50867570fd5040d91603950d` against merge-base `d7774ca` on `main`.

## Summary

Strong, well-disciplined Phase 4 implementation. The Kafka consumer (C4), reusable OpenSearch bulk writer, and embedded index mapping are correct, race-clean, and backed by comprehensive tests targeting exactly the failure modes that matter. **No P0/P1 blockers found.** Three P2 follow-ups below are suggestions, not merge blockers.

Verified locally: `go build ./...`, `go vet ./...`, and `go test -race ./...` all pass. Coverage: `internal/esindex` 85.4%, `internal/consumer` 67.7%. `gofmt` clean.

## Verification — what was checked and confirmed

- ✅ **Offset-commit correctness (no message loss).** `partitionCommitPoints` computes the contiguous success prefix *per partition*, truncating at the first transient; the in-place retry loop (`processBatch`) never fetches new offsets while a transient remains. Confirmed `kafka-go` `CommitMessages` internally commits `offset+1` (`commit.go: makeCommit`), so passing the last contiguous-success message correctly confirms `0..offset` inclusive. A high-watermark commit can never cross an unprocessed message. `consumer.go:121-168`, `decision.go:63-92`, `kafka.go:70-76`.
- ✅ **Transient vs permanent classification.** 429 correctly treated as transient; 4xx (non-429) permanent; 5xx / network / batch-level / missing-response-item all transient and never silently dropped. `writer.go:59-72,185-220`.
- ✅ **Idempotent sink.** `_id = message_id` via `index` action → duplicate delivery upserts the same doc (effectively-once over at-least-once). `writer.go:117-121,157`.
- ✅ **EnsureIndex TOCTOU under concurrent replica startup.** `isAlreadyExists` authoritatively requires a real HTTP 400 (`Inspect().Response.StatusCode`), not a body-claimed status — defends against a proxy returning HTTP 500 with `body status:400`. `existsNow` is the conservative fallback so a real 5xx with the index still absent propagates instead of being masked. Both landmines have dedicated tests (`TestEnsureIndex_HTTP500BodyClaims400Propagates`, `TestEnsureIndex_5xxWithTokenStillAbsentPropagates`). `ensure_index.go:43-108`.
- ✅ **Contract ↔ mapping alignment (cross-file).** Every `searchmsg.Message` JSON tag maps 1:1 to a property in `mapping/octo-message.json` under `dynamic: strict` — no field drift that would compile-but-silently-misindex. `SchemaVersion` mismatch routes to DLQ, never silently consumed.
- ✅ **DLQ terminal escape.** Bounded retry → spill-to-disk + alarm + advance (when spill dir set) or fatal hard-stop + page (when not). The prefix can never wedge forever, and a poison pill is never silently crossed. `dlq.go:79-134`.
- ✅ **Graceful shutdown / backoff.** `signal.NotifyContext` + ctx-aware `sleepCtx` means SIGTERM interrupts backoff immediately; exponential full-jitter mitigates multi-replica thundering herd. `main.go:38-46`, `backoff.go`.
- ✅ **Idle-by-default.** Binary no-ops unless `ES_INDEXER_ENABLED=true` AND brokers+ES configured — safe to deploy before enable. `main.go:73-92`.

## Findings (all P2 — non-blocking suggestions)

### P2-1 — Spill-to-disk durability assumption is undocumented
`dlq.go:113-134`. When `INDEXER_DLQ_SPILL_DIR` is set, `escape()` writes the record to local disk and returns `nil` so the offset advances. This is a sound *availability-over-durability* escape valve (it satisfies the C4 "never wedge forever" guarantee), but on a default Kubernetes pod the spill dir is ephemeral, so a crash/reschedule loses spilled records. The code is correct; the gap is documentation. Suggest the README explicitly state that spill mode requires a persistent volume (or knowingly accepts loss on pod churn), so operators mount a PVC.

### P2-2 — No recovery/replay path for spilled records
`service.go:96-101`, `main.go`. There is no startup phase that scans `INDEXER_DLQ_SPILL_DIR` and replays spilled records to the DLQ topic. Acceptable for a terminal escape hatch that should fire rarely, but pair it with P2-1: document the manual recovery procedure, or track replay tooling as a phase-7 operational follow-up. Records are preserved on disk (not silently lost), so this is tooling/runbook, not a correctness defect.

### P2-3 — `envInt()` silently falls back to default on parse error
`main.go:101-111`. A typo like `INDEXER_DLQ_MAX_RETRIES=invalid` logs at INFO and proceeds with the default. `BatchSize`/`TransientBackoff` get re-validated downstream (`<=0` reset in `NewProcessor`), but `DLQMaxRetries`/`DLQRetryBackoff` do not, so a mistyped value is swallowed. Given the codebase's fail-fast config philosophy, consider `log.Fatalf` (exit 1) on `strconv.Atoi` failure so misconfiguration surfaces at boot rather than degrading silently.

## Recommendation

**Approve.** Implementation is correct and the reliability contract (C4) holds. The three P2 items are documentation/operational hardening that can land as a follow-up; none block merge.

Cross-repo reminder (already noted in the PR): the octo-deployment managed-OpenSearch template must be provisioned at the same contract version (`searchmsg.SchemaVersion`) as the embedded `mapping/octo-message.json`; the `content.analyzer` override for IK/smartcn is owned by that deployment change.

PR #7, opened 2026-06-14
feat(harness): local e2e verification + reconciliation gate + tuning baseline

yujiawei:
## Summary

Phase 5: a self-contained **local** e2e verification harness, a reusable
**reconciliation** package/CLI (the phase-6 backfill correctness gate), and a
bulk-**tuning** baseline doc — all on a throwaway local stack.

> 🔒 **Isolation respected.** Everything here runs against a local `docker compose`
> stack only. No shared-environment `Kafka.On`, no octo-deployment wiring. Real
> deployment is a separate, explicitly signed-off step.

## (a) Local e2e harness (`harness/`)

`docker compose` brings up **Kafka (KRaft) + OpenSearch with the `analysis-ik`
plugin** (version-matched Dockerfile). `run.sh` builds/starts, pre-creates the
body + DLQ topics (the indexer's DLQ producer uses `AllowAutoTopicCreation=false`
by design), then runs the indexer, seeds, and verifies.

- `seed/` produces a controlled contract suite (normal / 中文 / `raw_excluded` /
  duplicate `_id` / unknown `schema_version` / bad JSON) and a `-mode bulk -n N` load.
- `verify/` asserts invariants against OpenSearch.

**Ran for real** (this PR was validated against the live stack):
- 9/9 invariants PASS through `searchetl`-shaped messages → Kafka → indexer → OpenSearch+IK.
- ~5,000 docs indexed end-to-end in ~1.4 s at batch 500; the 2 poison pills land
  in `octo.message.v1.dlq` (offset 2); IK segments `今天/北京/公园/散步` as CN words.

## (b) Invariants verified on the real stream

- **Idempotency** — duplicate `message_id` → exactly one ES doc.
- **C4** — unknown `schema_version` + bad JSON NOT indexed (→ DLQ); `raw_excluded`
  IS indexed (content null). A `consumer-drained` quiet-period gate runs before the
  negative assertions so they can't false-PASS.
- **IK tokenization** — Chinese (`公园`, `北京`) + English (`pipeline`) recall.

(C1/C3/C2 producer/consumer state-machine invariants remain unit-tested under
`-race`; the harness exercises the same code against real infra.)

## (c) Bulk tuning baseline (`docs/tuning.md`)

Recommended `INDEXER_BATCH_SIZE` (start 1000), transient/DLQ backoff, spill, and
OpenSearch-side guidance (`refresh_interval`, replicas, scale via partitions +
replicas), plus the phase-6 backfill starting config.

## (d) Reconciliation gate (`internal/recon` + `cmd/reconcile`)

The phase-6 correctness gate, **decoupled** from the writer (importable by the
backfill job):

```
ES_docs(window) == source_rows(window) - DLQ(window)
```

`raw_excluded` docs still occupy an ES doc (NOT subtracted); revoked/deleted keep
their doc too (route 甲 read-time join). MySQL per-shard `COUNT` (table-name
allowlist, no injection), OpenSearch `_count` (created_at range + raw_excluded
term). **The gate rejects partial-shard counts** (`_shards.failed`/`successful`)
so a degraded ES can't yield a false OK. CLI exits 2 on mismatch (vs 1 on error).
Validated against the live harness: `5007 source − 2 DLQ == 5005 ES docs → OK`.

## Mapping

`content` switched to **IK** (`ik_max_word` index / `ik_smart` search) per the
phase-4 tokenizer decision; the harness installs IK so the embedded mapping is
exercised for real. Deployment may override the analyzer if a cluster lacks IK.

## Checks

`go build` / `go vet` / `go test -race` / `golangci-lint` all clean. Independent
model review: PASS (after hardening the recon gate against partial-shard counts
and making the harness drain-gate fail-closed). Pure utility/harness commands have
no production runtime impact (the shipped image still builds only `./cmd/es-indexer`).

### DIFF
--- a/cmd/reconcile/main.go
+++ b/cmd/reconcile/main.go
@@ -1,0 +1,137 @@
+// Command reconcile 对账 MySQL message 表行数 vs OpenSearch doc 数（给定时间窗），扣除已知排除
+// 后输出差异报告。对平退出 0，不对平退出 2（可作为阶段 6 backfill 的正确性 gate / CI 检查）。
+//
+// 用法（env 或 flag 配置）：
+//
+//	reconcile -from 1709078400 -to 1718323200 \
+//	  -mysql-dsn 'user:pass@tcp(host:3306)/im_prod' -tables message,message1,message2,message3,message4 \
+//	  -es http://localhost:9200 -es-index octo-message [-dlq N]
+//
+// 与 indexer 写入器解耦（沿用 internal/esindex 解耦纪律）：本命令只读 MySQL + 查 ES，不依赖
+// consumer。阶段 6 backfill job 可直接 import internal/recon 复用对账算术。
+package main
+
+import (
+	"context"
+	"database/sql"
+	"flag"
+	"fmt"
+	"log"
+	"os"
+	"strings"
+	"time"
+
+	"github.com/Mininglamp-OSS/octo-search-indexer/internal/recon"
+	_ "github.com/go-sql-driver/mysql"
+	"github.com/opensearch-project/opensearch-go/v3"
+	"github.com/opensearch-project/opensearch-go/v3/opensearchapi"
+)
+
+func main() {
+	log.SetFlags(log.LstdFlags | log.LUTC)
+	if err := run(); err != nil {
+		log.Printf("reconcile error: %v", err)
+		os.Exit(1)
+	}
+}
+
+func run() error {
+	var (
+		fromUnix = flag.Int64("from", 0, "window start (epoch seconds, inclusive)")
+		toUnix   = flag.Int64("to", 0, "window end (epoch seconds, inclusive); 0 = now")
+		mysqlDSN = flag.String("mysql-dsn", os.Getenv("RECON_MYSQL_DSN"), "MySQL DSN (or RECON_MYSQL_DSN)")
+		tablesS  = flag.String("tables", envOr("RECON_TABLES", "message,message1,message2,message3,message4"), "comma-separated message shard tables")
+		esAddrs  = flag.String("es", envOr("RECON_ES", "http://localhost:9200"), "comma-separated OpenSearch addresses")
+		esIndex  = flag.String("es-index", envOr("RECON_ES_INDEX", "octo-message"), "OpenSearch index")
+		esUser   = flag.String("es-user", os.Getenv("RECON_ES_USER"), "OpenSearch username")
+		esPass   = flag.String("es-pass", os.Getenv("RECON_ES_PASS"), "OpenSearch password")
+		dlq      = flag.Int64("dlq", 0, "known DLQ count in window (rows that never reached ES body index)")
+		timeout  = flag.Duration("timeout", 60*time.Second, "overall timeout")
+	)
+	flag.Parse()
+
+	if *mysqlDSN == "" {
+		return fmt.Errorf("mysql DSN required (-mysql-dsn or RECON_MYSQL_DSN)")
+	}
+	to := *toUnix
+	if to == 0 {
+		to = time.Now().Unix()
+	}
+	if *fromUnix > to {
+		return fmt.Errorf("invalid window: from(%d) > to(%d)", *fromUnix, to)
+	}
+
+	ctx, cancel := context.WithTimeout(context.Background(), *timeout)
+	defer cancel()
+
+	db, err := sql.Open("mysql", *mysqlDSN)
+	if err != nil {
+		return fmt.Errorf("open mysql: %w", err)
+	}
+	defer func() {
+		if cerr := db.Close(); cerr != nil {
+			log.Printf("warn: close mysql: %v", cerr)
+		}
+	}()
+	if err := db.PingContext(ctx); err != nil {
+		return fmt.Errorf("ping mysql: %w", err)
+	}
+
+	osClient, err := opensearchapi.NewClient(opensearchapi.Config{
+		Client: opensearch.Config{
+			Addresses: splitCSV(*esAddrs),
+			Username:  *esUser,
+			Password:  *esPass,
+		},
+	})
+	if err != nil {
+		return fmt.Errorf("opensearch client: %w", err)
+	}
+
+	srcCounter := recon.NewMySQLSourceCounter(db, splitCSV(*tablesS))
+	esCounter := recon.NewOSCounter(osClient, *esIndex)
+
+	sourceRows, err := srcCounter.CountRows(ctx, *fromUnix, to)
+	if err != nil {
+		return err
+	}
+	esDocs, err := esCounter.CountDocs(ctx, *fromUnix, to)
+	if err != nil {
+		return err
+	}
+	rawExcluded, err := esCounter.CountRawExcluded(ctx, *fromUnix, to)
+	if err != nil {
+		return err
+	}
+
+	report := recon.Reconcile(recon.Counts{
+		SourceRows:  sourceRows,
+		ESDocs:      esDocs,
+		RawExcluded: rawExcluded,
+		DLQ:         *dlq,
+	})
+	fmt.Println(report.String())
+	if !report.OK {
+		// 不对平：退出码 2，作为 backfill/CI gate 的失败信号。
+		os.Exit(2)
+	}
+	return nil
+}
+
+func envOr(k, def string) string {
+	if v := os.Getenv(k); v != "" {
+		return v
+	}
+	return def
+}
+
+func splitCSV(s string) []string {
+	parts := strings.Split(s, ",")
+	out := make([]string, 0, len(parts))
+	for _, p := range parts {
+		if t := strings.TrimSpace(p); t != "" {
+			out = append(out, t)
+		}
+	}
+	return out
+}
--- a/docs/tuning.md
+++ b/docs/tuning.md
@@ -1,0 +1,57 @@
+# es-indexer bulk tuning — recommended baseline
+
+Recommendations from the local e2e harness (`harness/`), to be re-validated at
+deployment scale in phase 6. The harness is a single-node throwaway stack (Kafka
+KRaft single broker + OpenSearch single node with `analysis-ik`), so absolute
+throughput numbers are a **floor**, not a capacity plan — the value here is the
+relative trade-offs and a safe starting config.
+
+## What was measured
+
+- Pipeline: `seed → Kafka octo.message.v1 → es-indexer consumer → OpenSearch`.
+- Workload: 5,000 contract messages (~Chinese body, `ik_max_word` indexed).
+- Observed: ~5,000 docs indexed end-to-end in ~1.4 s at `INDEXER_BATCH_SIZE=500`
+  (≈3.5k docs/s) on the harness box, lag returning to 0, offset committed to the
+  log end, ES `_count` matching expected (5 suite + 5000 bulk = 5005, 2 poison
+  pills routed to the DLQ topic, 1 `raw_excluded` doc present).
+- Idempotency confirmed: duplicate `message_id` → exactly one ES doc.
+
+## Knobs (env on the indexer)
+
+| Env | Default | Meaning | Guidance |
+| --- | --- | --- | --- |
+| `INDEXER_BATCH_SIZE` | 500 | max docs per `_bulk` request | 500–2000. Larger = higher throughput + better amortised round-trips, but bigger bulk bodies and coarser retry granularity (a whole batch re-bulks its unresolved subset on transient). At ~700 B/doc (prod avg), 1000 docs ≈ 0.7 MB body — well within limits. **Start 1000.** |
+| `INDEXER_TRANSIENT_BACKOFF_MS` | 1000 | base for transient in-place retry (exp + full jitter, capped 30 s) | Keep ~1 s. Jitter already de-correlates replicas; raising it slows recovery, lowering it risks hammering a degraded ES. |
+| `INDEXER_DLQ_MAX_RETRIES` | 5 | DLQ-write retries before terminal escape | Keep 5. With exp backoff this is ~tens of seconds of tolerance for a DLQ blip before spill/hard-stop. |
+| `INDEXER_DLQ_RETRY_BACKOFF_MS` | 200 | base for DLQ-write retry backoff | Keep. |
+| `INDEXER_DLQ_SPILL_DIR` | "" | enable spill-to-disk terminal escape | **Set in production** to a durable, monitored path so a DLQ outage can't wedge the partition; leave empty only where a hard-stop+page is preferred. |
+
+## OpenSearch-side (owned by octo-deployment, not the indexer binary)
+
+| Setting | Recommendation | Why |
+| --- | --- | --- |
+| `index.refresh_interval` | `30s` (steady state); `-1` during phase-6 backfill, then restore + force `_refresh` | The indexer does not force per-batch refresh, so search visibility lags by `refresh_interval`. For bulk backfill, disabling refresh then re-enabling once is materially faster. |
+| `index.number_of_replicas` | `0` during backfill, raise to `1` after | Fewer replicas = faster bulk; restore for HA once loaded. |
+| `index.translog.durability` | `async` acceptable for a derived read model (SoR is MySQL) | The index is rebuildable from MySQL, so trading a tiny crash-window for throughput is safe. |
+| bulk concurrency | drive parallelism via **Kafka partitions + indexer replicas**, one consumer per partition | The writer issues one `_bulk` per batch synchronously; horizontal scale comes from partitions/replicas (each replica owns partitions, C3/C4 keep per-partition ordered commit correct), not in-process bulk fan-out. Keeps the failure model simple. |
+
+## Phase-6 backfill starting point
+
+- `INDEXER_BATCH_SIZE=1000`, ES `refresh_interval=-1`, `number_of_replicas=0`,
+  rate-limit ≤5k docs/s (matches the phase-6 plan), then restore index settings
+  and force one `_refresh`, then run the reconcile gate (below).
+- Backfill reuses `internal/esindex` (writer + `EnsureIndex` mapping) and
+  `internal/recon` (the gate) directly — no ES code is re-implemented.
+
+## Reconciliation gate
+
+`cmd/reconcile` (and `internal/recon`) implement the phase-6 correctness gate:
+
+```
+ES_docs(window) == source_rows(window) - DLQ(window)
+```
+
+`raw_excluded` docs (Signal/non-text) **still occupy an ES doc** (content null),
+so they are NOT subtracted. Revoked/deleted rows also keep their ES doc (route
+甲 filters at read time), so they are not subtracted either. Verified against the
+live harness OpenSearch: `5007 source - 2 DLQ == 5005 ES docs → OK (diff 0)`.
--- a/go.mod
+++ b/go.mod
@@ -4,6 +4,7 @@ go 1.25
 
 require (
 	github.com/Mininglamp-OSS/octo-lib v0.0.0-20260614033907-f62e626a34db
+	github.com/go-sql-driver/mysql v1.7.1
 	github.com/opensearch-project/opensearch-go/v3 v3.1.0
 	github.com/segmentio/kafka-go v0.4.51
 )
--- a/go.sum
+++ b/go.sum
@@ -2,6 +2,8 @@ github.com/Mininglamp-OSS/octo-lib v0.0.0-20260614033907-f62e626a34db h1:FudjIbM
 github.com/Mininglamp-OSS/octo-lib v0.0.0-20260614033907-f62e626a34db/go.mod h1:uOCTTI3QdG0ETKZ6uluRWUZgKexee/pY1pQYWyIEvZ0=
 github.com/davecgh/go-spew v1.1.1 h1:vj9j/u1bqnvCEfJOwUhtlOARqs3+rkHYY13jYWTU97c=
 github.com/davecgh/go-spew v1.1.1/go.mod h1:J7Y8YcW2NihsgmVo/mv3lAwl/skON4iLHjSsI+c5H38=
+github.com/go-sql-driver/mysql v1.7.1 h1:lUIinVbN1DY0xBg0eMOzmmtGoHwWBbvnWubQUrtU8EI=
+github.com/go-sql-driver/mysql v1.7.1/go.mod h1:OXbVy3sEdcQ2Doequ6Z5BW6fXNQTmx+9S1MCJN5yJMI=
 github.com/klauspost/compress v1.15.9 h1:wKRjX6JRtDdrE9qwa4b/Cip7ACOshUI4smpCQanqjSY=
 github.com/klauspost/compress v1.15.9/go.mod h1:PhcZ0MbTNciWF3rruxRgKxI5NkcHHrHUDtV4Yw2GlzU=
 github.com/opensearch-project/opensearch-go/v3 v3.1.0 h1:7EghS/+dCYD6PrsXjfIf3fvMOObkPtrDJVbovlNl3sY=
--- a/harness/README.md
+++ b/harness/README.md
@@ -1,0 +1,65 @@
+# Local e2e verification harness
+
+A throwaway local stack that runs the **whole** message-search pipeline for real:
+
+```
+seed → Kafka octo.message.v1 → es-indexer consumer → OpenSearch (analysis-ik)
+                                      └→ octo.message.v1.dlq (poison pills)
+```
+
+> 🔒 **Isolation.** This is a local-only stack (`docker compose`). It is NEVER
+> wired into a shared test environment, and the indexer's `Kafka.On` / real
+> deployment wiring stays off until a separate, explicitly signed-off step.
+> Real deployment provisioning (octo-deployment, IK dictionaries, topics) is out
+> of scope here.
+
+## Contents
+
+| Path | What |
+| --- | --- |
+| `docker-compose.yml` | Kafka (KRaft, 1 broker) + OpenSearch (1 node) |
+| `opensearch-ik.Dockerfile` | OpenSearch image with the `analysis-ik` plugin installed (version-matched) |
+| `run.sh` | orchestrator: `up` (build+start+create topics) / `down` / `seed` / `verify` / `all` |
+| `seed/` | producer of a controlled message suite (normal / 中文 / raw_excluded / duplicate _id / unknown schema / bad JSON) and a `-mode bulk -n N` throughput load |
+| `verify/` | asserts the invariants against OpenSearch |
+
+## Run it
+
+```sh
+# full cycle: up → indexer → seed suite → verify invariants → down
+./harness/run.sh
+
+# or step by step, leaving the stack up:
+KEEP_UP=1 ./harness/run.sh up
+ES_INDEXER_ENABLED=true KAFKA_BROKERS=localhost:19092 \
+  ES_ADDRESSES=http://localhost:19200 ES_INDEX=octo-message \
+  KAFKA_DLQ_TOPIC=octo.message.v1.dlq INDEXER_DLQ_SPILL_DIR=/tmp/spill \
+  go run ./cmd/es-indexer &        # start the indexer against the harness
+go run ./harness/seed -mode suite  # seed the controlled suite
+go run ./harness/verify            # assert invariants
+./harness/run.sh down
+```
+
+Ports: Kafka `localhost:19092`, OpenSearch `http://localhost:19200`.
+
+## Invariants asserted (`verify/`)
+
+- **Idempotency** — duplicate `message_id` → exactly one ES doc (`_id` upsert).
+- **C4 schema gate** — unknown `schema_version` and malformed JSON are NOT
+  indexed (routed to the DLQ topic).
+- **raw_excluded** — Signal/non-text doc IS indexed (content null), occupies a doc.
+- **IK tokenization** — Chinese terms (`公园`, `北京`) recall their docs; English
+  (`pipeline`) too.
+
+The C2/C4 producer/consumer state-machine invariants (per-partition ordered
+commit, transient in-place retry, DLQ terminal escape) are unit-tested under
+`-race` in `internal/consumer`; the harness exercises the same code paths against
+a real broker + real OpenSearch.
+
+## Notes
+
+- The DLQ topic must be **pre-created** (the indexer's DLQ producer uses
+  `AllowAutoTopicCreation=false` by design). `run.sh up` creates both topics; in
+  production the platform provisions them.
+- The harness OpenSearch disables the security plugin for plain-HTTP local
+  testing only — production uses authenticated endpoints (`ES_USERNAME`/`ES_PASSWORD`).
--- a/harness/README.md
+++ b/harness/README.md
@@ -35,18 +35,36 @@ ES_INDEXER_ENABLED=true KAFKA_BROKERS=localhost:19092 \
   ES_ADDRESSES=http://localhost:19200 ES_INDEX=octo-message \
   KAFKA_DLQ_TOPIC=octo.message.v1.dlq INDEXER_DLQ_SPILL_DIR=/tmp/spill \
   go run ./cmd/es-indexer &        # start the indexer against the harness
+
+# Fence the DLQ check to THIS run: capture the DLQ end offset BEFORE seeding.
+FENCE=$(docker exec octo-harness-kafka /opt/kafka/bin/kafka-get-offsets.sh \
+  --bootstrap-server localhost:9092 --topic octo.message.v1.dlq 2>/dev/null \
+  | awk -F: '{s+=$3} END{print s+0}')
 go run ./harness/seed -mode suite  # seed the controlled suite
-go run ./harness/verify            # assert invariants
+DLQ_START_OFFSET="$FENCE" go run ./harness/verify   # assert invariants
 ./harness/run.sh down
 ```
+
+> The verifier is **fail-closed** on the DLQ fence: if `DLQ_START_OFFSET` is not
+> provided and the DLQ topic already contains records, `verify` aborts rather than
+> risk matching stale poison-pill records from a previous run. On a fresh/empty DLQ
+> no fence is needed. `run.sh` captures and passes the fence automatically.
 
 Ports: Kafka `localhost:19092`, OpenSearch `http://localhost:19200`.
 
 ## Invariants asserted (`verify/`)
 
+- **Consumer-group drained (Kafka)** — the committed offset of the consumer group
+  reaches the body topic's log-end offset on every partition. This is the
+  authoritative drain proof (queried from Kafka directly, not inferred from ES doc
+  counts), so the negative "not indexed" assertions below are sound. It cannot be
+  fooled by a consumer that stalled before the poison pills or by the DLQ-spill path.
 - **Idempotency** — duplicate `message_id` → exactly one ES doc (`_id` upsert).
-- **C4 schema gate** — unknown `schema_version` and malformed JSON are NOT
-  indexed (routed to the DLQ topic).
+- **C4 schema gate — both directions:**
+  - positive: the **DLQ topic actually contains** the poison-pill keys
+    (`m-badschema`, `m-badjson`) — read straight from the DLQ topic, proving the
+    routing happened;
+  - negative: those messages are **NOT** in the ES body index.
 - **raw_excluded** — Signal/non-text doc IS indexed (content null), occupies a doc.
 - **IK tokenization** — Chinese terms (`公园`, `北京`) recall their docs; English
   (`pipeline`) too.
--- a/harness/docker-compose.yml
+++ b/harness/docker-compose.yml
@@ -1,0 +1,59 @@
+# Local e2e verification harness for the octo message-search pipeline.
+#
+# Kafka (KRaft, single broker) + OpenSearch (single node, analysis-ik installed).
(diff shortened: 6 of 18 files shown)


mochashanyao (approved):
[Octo-Q · automated review]

---

# octo-search-indexer PR#7 Review Report
**Head SHA:** `67f3bc9773beaba7751491f0d5756a7e04247cf6`
**PR:** feat(harness): local e2e verification + reconciliation gate + tuning baseline

---

## 1. Verification Summary

| Area | Status | Evidence |
|------|--------|----------|
| Reconciliation arithmetic | ✅ | `internal/recon/recon.go:57-66` — `Expected = SourceRows - DLQ` matches consumer data flow |
| Shard-safety gate | ✅ | `internal/recon/collectors.go:131-134` — rejects partial-shard counts; tested in `collectors_test.go` |
| Mapping change (cjk_bigram → IK) | ✅ | `internal/esindex/mapping/octo-message.json:18-19` — consistent with README + harness Dockerfile |
| Harness drain-gate (fail-closed) | ✅ | `harness/verify/main.go:72-82` — negative assertions gated on proven consumer drain |
| SQL injection protection | ✅ | `internal/recon/collectors.go:68-84` — `safeTableName` allowlist + parameterized time bounds |
| Idempotency (duplicate _id) | ✅ | `harness/seed/main.go:89-91` seeds dup; `esindex/writer.go:131` uses `index` action (upsert) |
| Test coverage | ✅ | recon: 7 tests; collectors: 4 tests; writer_test updated for IK |
| Isolation discipline | ✅ | No shared-env wiring; harness is docker-compose local only |

## 2. Findings

### P2-1: MySQL query uses `UNIX_TIMESTAMP(created_at)` — may prevent index range scan

**File:** `internal/recon/collectors.go:57`
**Severity:** P2 (performance, not correctness)
**Diff-scope:** New code (this PR introduces the recon package)

```go
q := fmt.Sprintf("SELECT COUNT(*) FROM `%s` WHERE UNIX_TIMESTAMP(created_at) BETWEEN ? AND ?", t)
```

Wrapping the column in `UNIX_TIMESTAMP()` prevents MySQL from using the `created_at` index for range scan on large production tables (full table scan per shard). The idiomatic form pushes the conversion to the parameter side:

```sql
WHERE created_at BETWEEN FROM_UNIXTIME(?) AND FROM_UNIXTIME(?)
```

**Recommendation:** Swap to `FROM_UNIXTIME` on the parameter side for phase-6 backfill-scale reconciliation runs. Acceptable as-is for initial harness validation.

### P2-2: DLQ count is manual CLI input — no auto-derivation from Kafka

**File:** `cmd/reconcile/main.go:52` (`-dlq` flag)
**Severity:** P2 (operational risk, not code defect)
**Diff-scope:** New code

The operator must supply the DLQ count for the window manually. If the count is wrong (too high or too low), the reconciliation produces a false OK or false MISMATCH. For phase-6 backfill this is a meaningful risk since the DLQ topic is the authoritative source.

**Recommendation:** For phase 6, consider auto-querying the DLQ topic's consumer lag or counting DLQ records by timestamp. Document the manual step clearly in the tuning doc.

### P2-3: `RawExcluded` collected but unused in reconciliation output

**File:** `internal/recon/recon.go:28` (field in `Counts`); `cmd/reconcile/main.go:107` (collected but not acted upon)
**Severity:** P2 (observability gap)
**Diff-scope:** New code

`CountRawExcluded` is called and the value is stored in `Counts.RawExcluded`, but the `Reconcile` function correctly does not subtract it (raw_excluded docs occupy ES docs). The `String()` output includes it parenthetically, which is good for observability. However, there's no cross-check: if `RawExcluded > ESDocs`, that would indicate a data anomaly worth flagging.

**Recommendation:** Consider adding a sanity warning when `RawExcluded > ESDocs` in the report output. Non-blocking.

### P2-4: Mapping change requires IK plugin — deployment coordination dependency

**File:** `internal/esindex/mapping/octo-message.json:18-19`
**Severity:** P2 (deployment risk, documented and mitigated)
**Diff-scope:** Modified (removes `octo_cjk_text` custom analyzer, adds `ik_max_word`/`ik_smart`)

The embedded mapping now requires the `analysis-ik` plugin. If `EnsureIndex` runs against a cluster without IK, index creation fails with a fatal error. This is:
- **Correctly documented** in the mapping README (override instructions for non-IK deployments)
- **Validated by the harness** (Dockerfile installs IK)
- **Not a regression for existing indexes** (`EnsureIndex` is no-op when index exists)

The risk is a new deployment that forgets to install IK or override the analyzer. The failure mode is loud (index creation error → service won't start), which is the correct fail-closed behavior.

**Recommendation:** Acceptable as-is. The coordinated deployment change owns the IK provisioning.

## 3. Recommendations

1. **Phase-6 prep:** Replace `UNIX_TIMESTAMP(created_at)` with `FROM_UNIXTIME` parameter conversion for index-friendly range scans at backfill scale.
2. **Phase-6 prep:** Auto-derive DLQ count from the Kafka DLQ topic (offset-based or record-count-by-timestamp) to eliminate manual input error.
3. **Minor:** Add a sanity cross-check for `RawExcluded > ESDocs` in the reconciliation report.

## 4. Additional Observations

- **Harness quality is high.** The drain-gate pattern (`waitDocCountStable` before negative assertions) correctly prevents false-PASS on poison-pill checks. The fail-closed design (drain unproven → record failure, not skip) is the right call.
- **Reconciliation package is cleanly decoupled.** `internal/recon` has zero imports from `internal/consumer` or `internal/esindex`, making it directly importable by the phase-6 backfill job as claimed.
- **go-sql-driver/mysql dependency** only affects `cmd/reconcile`, not the shipped `cmd/es-indexer` binary. No production runtime impact.
- **Test discipline is solid:** shard-safety gate has 3 dedicated tests (clean/partial-failure/incomplete); reconciliation arithmetic has 6 scenarios including edge cases.

## 5. Data Flow Trace

| Consumed Data | Upstream Source | Verified Flow |
|---|---|---|
| `SourceRows` (recon) | MySQL `COUNT(*)` per shard table, `created_at` range | ✅ `collectors.go:50-62` → parameterized query, safe table names |
| `ESDocs` (recon) | OpenSearch `_count` API, `created_at` range filter | ✅ `collectors.go:107-109` → `rangeQuery` → `count()` with shard gate |
| `RawExcluded` (recon) | OpenSearch `_count` with `raw_excluded:true` term filter | ✅ `collectors.go:112-125` → bool filter query |
| `DLQ` (recon) | Manual CLI input (`-dlq` flag) | ⚠️ Human-supplied, not auto-derived (P2-2) |
| `Expected` (recon) | `SourceRows - DLQ` | ✅ Matches consumer data flow: DLQ items never reach ES bulk |
| `ik_max_word` analyzer | Embedded mapping JSON | ✅ Loaded via `//go:embed` in `ensure_index.go:14`, sent on index create |
| Harness seed messages | `searchmsg.Message` contract structs → JSON → Kafka | ✅ `seed/main.go:78-100` uses `searchmsg.SchemaVersion` for valid msgs |
| Verify negative assertions | Gated on `waitDocCountStable` (8s quiet, 40s max) | ✅ `verify/main.go:72-82` fail-closed when drain unproven |

## 6. R5 Blind-Spot Checklist

- **C1 — Dual-path parity:** N/A. This PR adds new utility/harness code; no symmetric add/remove or subscribe/unsubscribe paths are modified.
- **C2 — Control-flow ordering / nested reuse:** N/A. The recon package is leaf-level (no emit/anchor/escape functions reused across call-sites). The harness is single-use orchestration.
- **C3 — Authorization boundary ≠ capability boundary:** N/A. The reconcile CLI is an operator tool (requires MySQL DSN + ES credentials). No new endpoints or tool exposure paths are created. The `go-sql-driver/mysql` dependency is confined to `cmd/reconcile`.

---

**[Octo-Q] verdict: APPROVE**

No P0 or P1 findings. All four P2 items are performance/operational improvements for phase-6 scale, not correctness blockers. The reconciliation arithmetic is verified correct against the consumer data flow. The shard-safety gate, drain-gate fail-closed design, and SQL injection protection are all solid. The IK mapping change is a documented deployment coordination item with correct fail-closed behavior.

Jerry-Xin (changes requested):
The PR is relevant to `octo-search-indexer`, but the new local e2e harness can falsely pass the poison-pill/DLQ invariant it claims to verify.

**🔴 Blocking**
- 🔴 Critical: [harness/verify/main.go](/Users/mlamp/.openclaw/tmp/pr-review-0my178d2/repo/harness/verify/main.go:66) treats a stable OpenSearch doc count as proof that the consumer drained, then only checks that bad messages are absent from ES. That does not prove the bad schema and malformed JSON records were consumed or routed to DLQ. If the indexer stops after indexing the positive cases, if it is stuck before the poison pills, or if DLQ topic creation fails and the configured spill path is used, the doc count can become stable and both negative assertions still pass. This directly contradicts the README claim that `verify/` asserts the C4 schema gate “routed to the DLQ topic” at [harness/README.md](/Users/mlamp/.openclaw/tmp/pr-review-0my178d2/repo/harness/README.md:47). Fix by making the verifier prove stream drain via Kafka state, such as committed group offset reaching the source topic log end, and assert the DLQ topic contains the two expected records or keys.

**💬 Non-blocking**
- 🟡 Warning: [harness/run.sh](/Users/mlamp/.openclaw/tmp/pr-review-0my178d2/repo/harness/run.sh:51) suppresses topic creation failures with `|| true`. That makes the above false-positive easier: a failed DLQ topic creation can be hidden, and because the harness sets `INDEXER_DLQ_SPILL_DIR` at [harness/run.sh](/Users/mlamp/.openclaw/tmp/pr-review-0my178d2/repo/harness/run.sh:75), the run may proceed without actually exercising Kafka DLQ delivery. It should ignore only “already exists” and fail on real creation errors.

**✅ Highlights**
- The reconciliation package is small and cleanly decoupled from the writer.
- The OpenSearch `_count` shard completeness check in [internal/recon/collectors.go](/Users/mlamp/.openclaw/tmp/pr-review-0my178d2/repo/internal/recon/collectors.go:130) is the right fail-closed behavior for a correctness gate.
- `go test ./...`, `go vet ./...`, and `go test -race ./internal/recon ./internal/esindex ./internal/consumer` all pass locally.

yujiawei (commented):
# Code Review — PR #7 (octo-search-indexer)

**Head SHA reviewed:** `67f3bc9773beaba7751491f0d5756a7e04247cf6`
**Base:** `main` (merge-base `c7f7712`) · **Scope:** +1271 / −21, 17 files

Phase-5: a local docker-compose e2e harness (`harness/`), a reconciliation
package + CLI (`internal/recon`, `cmd/reconcile`) for the phase-6 backfill gate,
a tuning baseline (`docs/tuning.md`), and a `content` analyzer switch from the
built-in `cjk_bigram` (`octo_cjk_text`) to the IK plugin (`ik_max_word` /
`ik_smart`).

## Verdict

**APPROVE** — no P0/P1 (correctness/security/data-loss/build-break) findings.
All findings below are P2/nit: they live in non-shipping code (the production
image still builds only `./cmd/es-indexer`; `harness/*` and `cmd/reconcile` are
not in it) or are phase-6 follow-ups, and none block merge. Build, vet, and the
full race test suite pass locally.

Note: this overlaps with a separate review that flagged the harness drain-gate.
I reach the same technical observation (item 1) but assess it as **P2, not
blocking**, for the reasons stated there.

## Verified ✅

- `go build ./...`, `go vet ./...`, `go test -race ./...` all clean (go 1.26).
- **Recon arithmetic** (`internal/recon/recon.go:57-64`): `Expected = SourceRows − DLQ`; `raw_excluded` correctly **not** subtracted (it still occupies an ES doc). Matches the consumer data flow and is covered by 6 table-tests incl. shortfall/surplus/DLQ.
- **Window alignment**: ES doc body is the marshaled source message, and `searchmsg.Message.CreatedAt == UNIX_TIMESTAMP(message.created_at)` (contract `searchmsg.go:61-62`). MySQL `BETWEEN ? AND ?` (inclusive) and ES `gte/lte` (inclusive) are symmetric — no boundary double-count.
- **Shard-safety gate** (`internal/recon/collectors.go:130-142`): `_count` with `_shards.failed != 0 || successful != total` is rejected. Correct fail-closed behavior for a correctness gate; 3 dedicated tests.
- **SQL-injection guard** (`collectors.go:48,66-92`): table names from a config allowlist (`safeTableName`, `[A-Za-z0-9_]` only) + parameterized time bounds. Tested with injection-style inputs.
- **Idempotency** (`esindex/writer.go:117-121,157`): `index` action with `_id = MessageID` → duplicate delivery upserts one doc. Seed exercises a duplicate `_id`.
- **Isolation**: harness is docker-compose local only; no shared-env `Kafka.On` wiring is introduced.

## Findings (P2 / nit — non-blocking)

### 1. (P2) Harness `verify` over-claims DLQ routing
`harness/verify/main.go:729-739` proves the poison pills are **absent from ES**
after a doc-count quiet period, but never confirms they **landed in the DLQ
topic**. `harness/README.md` and the PR body state the C4 gate is asserted as
"routed to the DLQ topic" — that specific claim is by inference, not measured.
A bug that *drops* poison pills entirely (neither ES nor DLQ) would still PASS.
The drain-gate also proves "no new docs," not "the consumer consumed offsets
6-7": a consumer hung exactly between the last positive (offset 5) and the first
poison pill could yield a stable count + a false PASS.
*Why non-blocking:* the C2/C4 DLQ state-machine is independently unit-tested
under `-race` in `internal/consumer`, and this is a local dev harness with zero
production runtime impact.
*Fix:* assert committed group offset == topic log-end, and consume
`octo.message.v1.dlq` to confirm the 2 expected records/keys are present.

### 2. (P2) `EnsureIndex` never updates an existing index → silent analyzer drift after the cjk→IK switch
`internal/esindex/ensure_index.go:44-48` short-circuits (`return nil`) when the
index already exists; the new IK mapping is only ever applied at first-create
(`:55-58`). A cluster that already bootstrapped `octo-message` under the old
`octo_cjk_text` analyzer keeps tokenizing with `cjk_bigram` while code/docs/tests
all assert IK — search behavior silently diverges from the contract. The
no-overwrite behavior is intentional (PR#6 TOCTOU safety) and likely fine for a
pre-GA index, but it is **undocumented**.
*Fix:* add one line to `internal/esindex/mapping/README.md` stating that analyzer
changes here do not propagate to existing indices and require a deployment-side
reindex (alias swap / close-create-reopen).

### 3. (P2) IK mapping is a hard cross-repo dependency; the zero-dependency fallback was removed
`internal/esindex/mapping/octo-message.json:18-19` hardcodes `ik_max_word`/
`ik_smart` and deletes the built-in `octo_cjk_text` analysis block.
`EnsureIndex` sends the mapping verbatim; on a cluster without `analysis-ik`,
`Indices.Create` fails with a non-`already_exists` error → `EnsureIndex` returns
error → es-indexer crashes on startup (CrashLoopBackOff). The override lives only
in the deployment repo.
*Why non-blocking:* documented, fail-loud, and the deployment change owns IK
provisioning in lockstep.
*Fix:* make IK a verifiable deploy precondition (assert `_cat/plugins` lists
`analysis-ik` before rollout), and/or add a startup pre-flight that emits an
explicit "analysis-ik plugin missing" fatal rather than a raw create-index error.

### 4. (P2) Reconcile `-dlq` is operator-supplied, not measured
`cmd/reconcile/main.go:54,111` feeds a manual `-dlq` flag straight into the gate
equation; there is no DLQ collector (unlike the MySQL/ES collectors). A wrong
value yields a false OK/MISMATCH. Default `-dlq 0` fails **safe** (any real DLQ
rows surface as MISMATCH), and this CLI is not yet wired into phase-6, so it is a
v1 limitation — but a "correctness gate" with one un-measured term is worth a
follow-up.
*Fix:* before phase-6 runs this unattended, add a DLQ collector (topic depth /
indexer-emitted counter); interim, WARN when `-dlq` is unset.

### 5. (P2) Recon against a live window can spuriously MISMATCH; doc lacks the caveat
`cmd/reconcile/main.go` defaults `to` to `time.Now()` and counts in one shot with
no settle. ES `_count` only sees refreshed docs and `docs/tuning.md:33` notes
visibility lags by `refresh_interval` (~30s). Run against live traffic, in-flight
rows are in MySQL but not yet ES-visible → `Diff<0` → `os.Exit(2)`. The phase-6
flow (quiesce + force `_refresh`) avoids this, but the tuning doc never states
recon requires a quiesced index.
*Fix:* document that recon must run against a quiesced/force-refreshed index (or
`-to` set safely in the past); optionally default `to` to `now − margin`.

### 6. (P2) Empty `-tables` makes the gate report a false OK
`internal/recon/collectors.go:48-62`: `CountRows` over an empty slice returns
`(0, nil)`. With `-tables ""` (or empty `RECON_TABLES`) and an empty ES window,
`expected=0, diff=0, OK=true, exit 0` — a misconfiguration reports green from a
gate whose whole purpose is to fail loudly.
*Fix:* reject an empty tables slice in `run()`/`NewMySQLSourceCounter`.

### 7. (P2) `internal/recon/recon.go:8-11` header invariant contradicts the implementation
The package-doc invariant reads `ES(...) + DLQ + 已知排除(raw_excluded) == 源行数`
(i.e. raw_excluded subtracted), but the code computes `Expected = SourceRows − DLQ`
and the rest of the file (and `TestReconcile_RawExcludedStillCounts`) correctly
states raw_excluded is **not** subtracted. A phase-6 author reading only the
header could double-count. *Fix:* rewrite the header formula to `ESDocs ==
SourceRows − DLQ`.

### 8. (P2) `run.sh stop_indexer` kills the subshell, not the indexer
`harness/run.sh:454-467,471-476`: `( ... go run ./cmd/es-indexer ) &` then `$!`
captures the **subshell** PID; `kill` does not stop `go run` or the compiled
binary. Harmless in the default `all` flow (`down` tears the stack), but under
`KEEP_UP=1` the indexer is orphaned, still joined to the consumer group and racing
later `seed`/`verify`. *Fix:* kill the process group (`setsid`/`kill -- -$PID`) or
`pkill -f 'go run ./cmd/es-indexer'`.

### 9. (P2) Non-sargable recon query
`internal/recon/collectors.go:48`: `WHERE UNIX_TIMESTAMP(created_at) BETWEEN ? AND ?`
wraps the indexed column → full scan per shard; at prod scale this can exceed the
single 60s context (`cmd/reconcile/main.go:55`). Correct, just slow; fails safe.
*Fix:* `created_at BETWEEN FROM_UNIXTIME(?) AND FROM_UNIXTIME(?)` and/or a larger
timeout for backfill-scale runs.

### Nits
- `cmd/reconcile/main.go:122` `os.Exit(2)` inside `run()` skips the deferred `cancel()`/`db.Close()`; return a code to `main()` instead. Harmless for a one-shot CLI.
- `harness/verify/main.go`: per-check 45s wall-clock poll deadlines are decoupled from the 90s parent context; a slow run spins with confusing "returned 0" errors rather than a clean timeout (still a FAIL, never a false PASS).
- `harness/run.sh` `trap 'stop_indexer' EXIT` does not `down` on a failed `seed`/`verify` under `set -e`; the stack is left up (self-heals next run).
- `harness/opensearch-ik.Dockerfile:8-9` installs IK from a mutable `/stable/` URL with no checksum/pin; fine for a throwaway local image, worth pinning if it ever feeds CI.

## Recommendation

Mergeable as-is. Highest-value follow-ups: item 1 (assert DLQ presence + offset
in the harness so its README claim is true) and item 2 (document the
reindex requirement for the analyzer switch). Items 4-7 should land before phase-6
relies on the recon gate unattended.

lml2468:
## ✅ Re-review — octo-search-indexer PR #7 (3f71bdb): harness false-PASS fixed, verified — approving

Re-cloned at 3f71bdb. Steve's 🔴 (the reconciliation gate could false-PASS the poison-pill→DLQ invariant it claims to verify) is fixed, more rigorously than the minimum. My initial review missed this — I checked the reconcile SQL but didn't audit the verify-harness assertions; Steve caught it.

### 🔴→✅ Authoritative drain + positive DLQ assertion
1. Drain proof now via Kafka, not ES doc-count (verify/main.go:79-85): expectGroupDrained requires the consumer group's committed offset to reach the body topic's log-end on every partition — closes the "indexer stalls before the pills → ES count settles → false-PASS" hole.
2. Positive DLQ assertion (:102): reads the DLQ topic and asserts the poison-pill keys (m-badschema, m-badjson) are actually present (routing happened), instead of the old "absent from ES ⇒ pass" negative-only check.
3. Fence rigor, fail-closed (:43-59): if -dlq-start-offset isn't given and the DLQ is non-empty, verify refuses to run rather than judge against stale records from a prior run; run.sh captures the DLQ end offset before seeding. Right way to keep a verification gate trustworthy — fails closed on ambiguity instead of false-passing.

### 🟡→✅ Topic creation
DLQ topic must be pre-created (indexer's DLQ producer uses AllowAutoTopicCreation=false); run.sh up creates both topics (README:79-80). Topic existence is now an explicit precondition.

### Verified
Build OK; go test ./... green (recon tests pass). The harness now proves what it claims: stream fully drained (Kafka offset) + poison pills actually in the DLQ topic.

### Carried non-blocking 🟡
reconcile UNIX_TIMESTAMP(created_at) non-sargable (one-shot ops cmd) — worth the FROM_UNIXTIME one-liner; DLQ manual count; RawExcluded anomaly signal.

LGTM.

lml2468 (approved):
Harness false-PASS fixed at 3f71bdb: drain proven via Kafka group-offset==log-end (not ES doc-count), positive DLQ assertion reads the DLQ topic to confirm poison-pill keys present, fail-closed fencing refuses stale DLQ records. Build+tests green. Gate now actually verifies poison→DLQ. My initial review missed this assertion gap — Steve caught it. LGTM.

mochashanyao (approved):
[Octo-Q · automated review]

---

# octo-search-indexer PR#7 Review Report
**Head SHA:** `3f71bdb6cdad091316968191d776caf2e92515fb`
**PR:** feat(harness): local e2e verification + reconciliation gate + tuning baseline

---

## 1. Verification Summary

| Area | Status | Evidence |
|------|--------|----------|
| Reconciliation arithmetic | ✅ | `internal/recon/recon.go:57-66` — `Expected = SourceRows - DLQ`, `Diff = ESDocs - Expected`. Correct per design: raw_excluded still occupies an ES doc, revoked/deleted keep their doc (route 甲 read-time join). |
| Partial shard failure gate | ✅ | `internal/recon/collectors.go:131-134` — rejects count when `Shards.Failed != 0 || Successful != Total`. Test coverage in `collectors_test.go:55-73`. |
| Consumer drain proof (Kafka-native) | ✅ | `harness/verify/kafka.go:42-72` — committed offset vs log-end offset per partition. Cannot be fooled by stalled consumer or DLQ-spill path. |
| DLQ positive verification | ✅ | `harness/verify/kafka.go:148-168` — reads DLQ topic from fence offset, asserts poison-pill keys actually present. Fail-closed on stale records. |
| IK mapping change | ✅ | `internal/esindex/mapping/octo-message.json:17-18` — `ik_max_word` index / `ik_smart` search. Test updated in `writer_test.go:240`. |
| Table name injection prevention | ✅ | `internal/recon/collectors.go:63-86` — `safeTableName` allows only `[A-Za-z0-9_]`. Parameters use SQL placeholders. |
| Test coverage (recon) | ✅ | `recon_test.go` covers all-indexed, DLQ-excluded, raw_excluded-still-counts, shortfall, surplus, combined. `collectors_test.go` covers clean/partial-shard/incomplete/raw_excluded. |
| Harness isolation | ✅ | No shared-environment wiring. `docker-compose.yml` local-only. `run.sh` pre-creates topics with proper error handling (commit 2 fix). |

## 2. Findings

### F1 — IK plugin hard dependency in embedded mapping (P2)

**Diff-scope:** NEW (this PR introduces the IK dependency into the embedded mapping).

`internal/esindex/mapping/octo-message.json` removes the self-contained `octo_cjk_text` analyzer and replaces it with `ik_max_word` / `ik_smart`, which require the `analysis-ik` plugin on the OpenSearch cluster. The `analysis` settings block is removed entirely.

**Data flow trace:** `EnsureIndex` (`ensure_index.go:47`) reads the embedded `mappingJSON` and passes it directly to `Indices.Create`. If the cluster lacks the IK plugin, the Create call will fail with an analyzer-not-found error. There is no code-level override mechanism — the README says "deployment may override" but that requires pre-creating the index with a different mapping before the indexer starts.

**Severity rationale (R1/R2):** This is a documented design decision, not a bug. The PR explicitly scopes itself to a local harness ("Real deployment is a separate, explicitly-signed-off step"). The mapping change is intentional and the deployment path would need to ensure IK is available or pre-create the index. No existing production index is affected because `EnsureIndex` is idempotent (won't overwrite an existing mapping). P2 — not a blocker.

### F2 — `os.Exit(2)` bypasses deferred `db.Close()` in reconcile CLI (P2/nit)

**Diff-scope:** NEW.

`cmd/reconcile/main.go:119` calls `os.Exit(2)` when `!report.OK`. The `defer db.Close()` at line 82 will not execute because `os.Exit` does not run deferred functions.

**Severity rationale:** The process is terminating; the OS reclaims the TCP connection. Acceptable for a CLI gate tool. P2/nit — cosmetic. If desired, restructure to return a sentinel error and let `main()` handle the exit code.

### F3 — `splitCSV` / `envOr` duplicated across three CLI commands (nit)

**Diff-scope:** NEW.

`splitCSV` and `envOr` are independently defined in `cmd/reconcile/main.go`, `harness/seed/main.go`, and `harness/verify/main.go`.

**Severity rationale:** These are separate `main` packages (not library code). The duplication is minor and intentional (each is a standalone binary). nit — no action required.

### F4 — MySQL `UNIX_TIMESTAMP(created_at)` timezone sensitivity (P2)

**Diff-scope:** NEW (recon tool).

`internal/recon/collectors.go:56` uses `UNIX_TIMESTAMP(created_at) BETWEEN ? AND ?`. MySQL's `UNIX_TIMESTAMP` interprets `DATETIME` columns in the server's session timezone. If the MySQL server timezone differs from UTC (which the ES `created_at` long field assumes), the reconciliation window may be offset by hours.

**Severity rationale:** A timezone mismatch would cause a false MISMATCH (the safe direction — operator investigates and adjusts), never a false OK. The recon tool is a correctness gate, not a production service. P2 — document the timezone assumption or add a `-tz` flag if this becomes operationally painful.

### F5 — `readPartitionKeys` 15-second per-partition timeout (nit)

**Diff-scope:** NEW (harness).

`harness/verify/kafka.go:222` sets a 15-second context timeout for reading a single partition. For the single-partition local harness this is generous. On a multi-partition topic with many records, it could theoretically be tight.

**Severity rationale:** Test harness only, not production code. nit.

## 3. Suggestions

- **F2**: Consider returning a typed error from `run()` and mapping it to exit code 2 in `main()`, so deferred cleanup runs normally. Low priority.
- **F4**: Consider documenting the timezone expectation in `docs/tuning.md` or adding a `-mysql-tz` flag to the reconcile CLI for operational flexibility.

## 4. Additional Observations

- The commit-2 fix (prove drain via Kafka offsets + positively verify DLQ routing) is a significant improvement over the original "ES doc count is stable" approach. The negative control (pointing DLQ check at wrong topic fails) validates the check has no false-positive.
- The `topicEndOffset` function (`kafka.go:260-280`) correctly handles the `UnknownTopicOrPartition` error as "topic not yet created → 0" and propagates all other errors, ensuring fail-closed behavior.
- The `create_topics` function in `run.sh` (lines 56-76) properly distinguishes "already exists" from genuine failures — a good hardening from the original `|| true` pattern.

## 5. Data Flow Trace

|---|---|---|
| `resp.Count` in `OSCounter.count` | OpenSearch `_count` API response | ✅ — gated by shard completeness check before return |
| `resp.Shards.Failed/Successful/Total` | OpenSearch `_count` API `_shards` field | ✅ — checked at `collectors.go:131-134` |
| MySQL `COUNT(*)` per shard table | `database/sql` QueryRowContext with placeholder params | ✅ — table name validated, time params bound |
| `searchmsg.Message` fields in seed | Constructed in `suite()` / `bulk()` with explicit field values | ✅ — all fields set, `json.Marshal` serializes per contract |
| Kafka committed offset in verifier | `OffsetFetch` API → `p.CommittedOffset` | ✅ — `-1` (no commit) excluded from map, `allCaughtUp` fails on missing |
| DLQ record keys in verifier | `kafka.NewReader` per-partition read from `startOffset` to `logEnd` | ✅ — fenced by `startOffset`, terminates at `end-1` |
| Embedded mapping in `EnsureIndex` | `//go:embed mapping/octo-message.json` | ✅ — compiled into binary, IK analyzers resolved at index-create time by OpenSearch |

## 6. R5 Blindspot Checklist

- **C1 (dual-path parity):** N/A — this PR adds test infrastructure and a read-only reconciliation tool. No add/remove, subscribe/unsubscribe, or create/delete symmetry paths are touched.
- **C2 (control-flow ordering / nested reuse):** N/A — no shared utility functions are called from multiple sites with different orderings. The `splitCSV`/`envOr` duplication is independent copies, not shared.
- **C3 (authorization boundary ≠ capability boundary):** N/A — no authorization or capability changes. The reconcile CLI takes explicit credentials via flags/env. The harness is local-only with security plugin disabled.

---

**[Octo-Q] verdict: APPROVE** — No P0 or P1 findings. All issues are P2/nit level (IK dependency documented, os.Exit cosmetic, code duplication minor, timezone assumption safe-direction). The PR is well-structured with strong defensive coding (partial shard gate, Kafka-native drain proof, DLQ fence fail-closed) and good test coverage.

Jerry-Xin (approved):
Summary: The PR is relevant to `octo-search-indexer` and adds useful local e2e coverage plus a reusable reconciliation gate; I found no merge-blocking defects.

💬 Non-blocking

🟡 Warning — `internal/recon/collectors.go:47-48`: `UNIX_TIMESTAMP(created_at)` makes the MySQL count non-sargable and may force full scans across all message shards during a production-scale backfill gate. It also couples correctness to the MySQL session timezone if `created_at` is `DATETIME`. Prefer converting the input bounds once and querying the column directly, or using an indexed/generated epoch column.

🟡 Warning — `cmd/reconcile/main.go:91` + `internal/recon/collectors.go:41-56`: an empty `-tables` value silently produces `source_rows=0`, which can yield a misleading OK on an empty ES window. Since this is a correctness gate, validate that `splitCSV(*tablesS)` is non-empty and fail closed.

🔵 Suggestion — `cmd/reconcile/main.go:102-110` + `internal/recon/recon.go:26-28`: `RawExcluded` is documented as a source-row count, but the CLI fills it from OpenSearch. It is currently observational only, so this is not a correctness bug, but either rename it to `ESRawExcluded` or add a true source-side count to avoid confusion during incident triage.

🔵 Suggestion — `cmd/reconcile/main.go:114-116`: calling `os.Exit(2)` inside `run()` bypasses deferred cleanup. Return a sentinel mismatch error/status from `run()` and let `main()` choose exit code `2`.

✅ Highlights

- The PR is in scope for this repository: it extends the indexer’s local verification, mapping, and backfill reconciliation workflow.
- Good fail-closed behavior in the harness DLQ fence and Kafka drain proof.
- Good OpenSearch `_count` hardening against partial shard results in `internal/recon/collectors.go:130-135`.
- `go test ./...` and `go vet ./...` both pass locally.

yujiawei (commented):
# Code Review — PR #7 (octo-search-indexer)

**Head SHA reviewed:** `3f71bdb6cdad091316968191d776caf2e92515fb`
**Base:** `main` (merge-base `c7f7712`) · **Scope:** +1629 / −21, 18 files

Independent re-review of the current head. The earlier blocking finding on the
e2e harness (the verifier proving consumer drain via a *stable OpenSearch doc
count*, which could false-PASS the poison-pill/DLQ invariant) has been
**genuinely fixed** in commit `3f71bdb` — see verification below. `go build`,
`go vet`, and `go test -race ./...` all pass locally; mapping/analyzer change is
consistent across code, embedded JSON, and tests.

## 1. Verification of the prior blocking finding (now resolved ✅)

- ✅ **Authoritative drain proof via Kafka, not ES doc-count.**
  `verify/main.go:84` now calls `expectGroupDrained`, which asserts committed
  offset ≥ log-end offset on every partition (`kafka.go:36-82`). A consumer that
  stalled *before* the poison pills, or one that took the DLQ-spill path, leaves
  a committed offset short of the log end → drain fails. The negative
  "not-indexed" assertions are gated behind `drainErr == nil`
  (`verify/main.go:106-112`), so they can no longer false-PASS.
- ✅ **Positive DLQ routing assertion.** `verify/main.go:105` now asserts the DLQ
  topic actually *contains* the two poison-pill keys (`m-badschema`, `m-badjson`)
  via `expectDLQKeys` (`kafka.go:140-157`), not merely their absence from ES.
- ✅ **Fail-closed DLQ fence.** When no `-dlq-start-offset` is supplied and the
  DLQ is non-empty, the verifier aborts rather than risk a stale-record false
  PASS (`verify/main.go:50-62`); `run.sh` captures the pre-seed offset
  (`run.sh:116-118`).
- ✅ **Prior 🟡 (`|| true` suppressing topic-creation failures) fixed.**
  `run.sh:51-71` now tolerates only "already exists" and fails loudly on any
  other non-zero rc.
- ✅ **Partial-shard fail-closed gate** in `internal/recon/collectors.go:130-135`
  is the correct behavior for a correctness gate, with test coverage
  (`collectors_test.go:50-67`).

## 2. Findings

### P1 — `reconcile`'s DLQ term is operator-supplied; the gate can fail-*open*
`cmd/reconcile/main.go:48,111` · `internal/recon/recon.go:57`

The gate computes `expected = SourceRows − DLQ`, then `OK = (ESDocs == expected)`.
`SourceRows` is measured live from MySQL and `ESDocs` from OpenSearch, but `DLQ`
is a manual `-dlq` flag (default 0) — the only input that is *not* independently
measured. The error direction is asymmetric:

- Omitting `-dlq` when DLQ rows exist → `diff < 0` → MISMATCH (**fail-safe**).
- Passing an inflated `-dlq` (e.g. a stale value baked into a wrapper script) can
  cancel a genuine ES shortfall to `diff == 0` → **false OK**, masking real data
  loss — exactly what this gate exists to catch.

This does not affect any production runtime path today (the reconcile CLI is a
phase-6 utility, not wired into the shipped image, which still builds only
`./cmd/es-indexer`), and the safe default (DLQ=0) is fail-safe — so it is not a
blocker for landing this scaffold. **But before the gate is actually wired into
the phase-6 backfill, the DLQ count should be measured, not trusted:** the
machinery already exists in this PR (`harness/verify/kafka.go` reads a Kafka
topic to its per-partition log-end), or count a DLQ landing table. At minimum,
gate mode should refuse a bare/defaulted `-dlq`.

### P2 — Package-doc invariant contradicts the implementation
`internal/recon/recon.go:9-11`

The package header states the invariant as
`ES_docs + DLQ + raw_excluded == SourceRows` (i.e. `raw_excluded` *subtracted*
from expected). The implementation, the function-level comment (`recon.go:46-55`),
and the tests (`TestReconcile_RawExcludedStillCounts`, `recon_test.go:26-32`) all
correctly do the opposite — `raw_excluded` occupies an ES doc and is **not**
subtracted (`expected = SourceRows − DLQ`). The code is correct; the authoritative
package doc is wrong and, as the phase-6 gate's contract doc, could mislead the
phase-6 implementer into subtracting `raw_excluded` and breaking the gate. Fix the
header to `ES_docs + DLQ == SourceRows`.

### Nits (non-blocking)

- **DLQ fence uses a cross-partition *sum* as a per-partition floor**
  (`harness/verify/kafka.go:163,175,215`; sum produced at `run.sh:76-80`). Correct
  today because the harness pins the DLQ to `--partitions 1` (`run.sh:59`), so
  sum == the single partition's log-end. It would silently *false-FAIL* DLQ
  verification if the DLQ topic were ever given >1 partition. The `kafka.go:159-162`
  comment calls this "conservative" — it actually over-fences; consider a hard
  assert that the DLQ is single-partition, or a true per-partition floor.
- **`readPartitionKeys` discards accumulated keys on a mid-read 15s timeout**
  (`kafka.go:208-221`), re-reading the partition from scratch on retry. Harmless
  within the 60s outer budget and correctly non-fatal; just suboptimal.
- **`os.Exit(2)` on mismatch bypasses the deferred `db.Close()`**
  (`cmd/reconcile/main.go:116`). No real leak (process exits, OS reclaims), but
  mixing `os.Exit` with `defer` is an anti-pattern — prefer returning a
  sentinel error and exiting from `main`.

## 3. Recommendations

1. Before phase-6 wires `reconcile` into the backfill, make the DLQ term measured
   (Kafka DLQ topic count / landing table) and reject a defaulted `-dlq` in gate
   mode (P1 above).
2. Correct the `internal/recon/recon.go` package-header invariant (P2 above).
3. Optionally pin/assert DLQ single-partition in the harness verifier (nit).

## 4. Additional notes

- **IK mapping migration is internally consistent**: `octo-message.json` switched
  `content` to `ik_max_word`/`ik_smart`, the embedded mapping is what `EnsureIndex`
  ships (`internal/esindex/ensure_index.go:18-19,55-58`), `writer_test.go:240` was
  updated to assert `ik_max_word`, and no stale `octo_cjk_text` references remain
  except an intentional README example. The deployment-ordering caveat (a cluster
  without `analysis-ik` must pre-create the index with an overridden analyzer,
  since `EnsureIndex` never overwrites an existing index) is owned by the
  octo-deployment change and noted in `mapping/README.md:27-30` — acceptable as
  documented, worth keeping visible in the deployment runbook.
- Isolation discipline holds: the harness is a throwaway local `docker compose`
  stack only; the shipped image builds just `./cmd/es-indexer`.

**Verdict:** No P0/P1 *blocks landing this scaffold* — the prior P0 is fixed, all
checks are green, and the one P1-level concern is a hardening requirement for a
phase-6 gate that is not yet operational (safe default holds, zero production
runtime impact). Approving, with the recon DLQ hardening (P1) and the doc-invariant
correction (P2) tracked as required before the gate goes live in phase-6.

PR #8, opened 2026-06-14
feat(backfill): historical message backfill job + DLQ-accounting reconcile gate

yujiawei:
## What

Phase 6 of the message-search delivery: a **historical backfill** job that loads the
existing `message` shard tables into OpenSearch, **bypassing Kafka**, reusing the same
write path as the live indexer.

```
message 5 shards → [cmd/backfill: keyset scan → internal/esindex.Writer bulk] → OpenSearch
```

The live Kafka increment keeps flowing unchanged; the two paths are safe to run in
parallel because the ES doc `_id = message_id` makes every write an idempotent upsert.

## Highlights

- **Reuses `internal/esindex`** (writer + `EnsureIndex` IK mapping + `_id=message_id`) —
  no ES write code re-implemented.
- **Keyset pagination** by PK (`WHERE id>? ORDER BY id ASC`), never `OFFSET`.
- **Idempotent** with the live increment; **resumable** per-shard checkpoint (atomic
  write, isolated from the live cursor); **rate-limited** (default ≤5k docs/s).
- **payload extraction byte-aligned with the live producer** (P1-d three-way split),
  pinned by a test that mirrors the producer's vectors + the two octo-lib constants.
- **DLQ accounting for the bypass-Kafka path**: a durable local spill file is the
  dedup'd source of truth (deduped by `message_id`, replayed/counted on reopen,
  fsynced before the checkpoint advances, directory-fsynced on creation), with a
  windowed count for the reconcile gate. `fail-closed` on any write failure.
- **Reconcile gate hardened** (`ReconcileChecked`): rejects self-inconsistent inputs
  (e.g. `DLQ > source_rows`, `raw_excluded > es_docs`) so DLQ accounting can't mask a
  shortfall as a false OK. The inline gate force-refreshes ES before counting.

Equation: `ES_docs(window) == source_rows(window) − DLQ(window)`. `raw_excluded`
(Signal/non-text) and revoked/deleted rows keep their ES doc and are not subtracted.

## Verified end-to-end (real MySQL + OpenSearch with IK)

`harness/run-backfill.sh` seeds the shard tables, runs the backfill, the inline
reconcile gate, and asserts the result. Observed: 6 source rows → 5 ES docs,
2 `raw_excluded`, 1 real anomaly spilled to DLQ and **positively absent** from ES;
`reconcile OK | diff 0`; re-run is a no-op via the checkpoint (idempotent, ES count
stays 5); IK 中文 recall works on backfilled docs (公园 → the 公园 row, 北京 → the 北京 row).

## Scope / safety

- New tool binary `cmd/backfill` + harness are **not** baked into the `es-indexer`
  service image (same precedent as `cmd/reconcile`); the production Dockerfile is
  unchanged. No new dependencies.
- Running against a **real** environment is a separate, explicitly signed-off,
  low-peak ops action — never automatic. The job only **reads** the message tables.

## Review

- /review (self): SQL safety (allowlisted table names on every dynamic name incl. the
  harness DROP/CREATE), conditional side effects, fail-closed gates — PASS ✅
- /codex review: PASS ✅ after 3 convergent rounds — all findings were DLQ-durability
  (replay count on reopen, idempotent writes, windowed count, fsync-before-checkpoint,
  directory-entry fsync), each fixed and covered by tests.
- `go build` / `go vet` / `go test -race` / `golangci-lint` all green.

See `docs/backfill.md` for the runbook and `harness/README.md` for the e2e.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,6 +9,11 @@ todos
 # Test binary
 *.test
 
+# Ad-hoc local build artifacts (e.g. `go build -o backfill ./cmd/backfill`)
+/backfill
+/es-indexer
+/reconcile
+
 # Output
 *.out
 
--- a/cmd/backfill/main.go
+++ b/cmd/backfill/main.go
@@ -1,0 +1,265 @@
+// Command backfill 是 octo-im 消息检索管线的历史回灌作业（YUJ-4534 阶段 6）。
+//
+// 它读 MySQL message 5 分表 → 复用 internal/esindex 写入器（同一套 bulk / mapping /
+// `_id=message_id`）直接幂等 bulk 写 OpenSearch，**绕开 Kafka**。与实时增量（Kafka →
+// es-indexer）并行 / 重叠安全：`_id=message_id` 幂等覆盖，同一条消息被两条路写入即覆盖同一 doc。
+//
+//	message 5 分表 → 【backfill 本作业：keyset 扫描 → esindex.Writer bulk】 → OpenSearch
+//
+// 设计纪律：
+//   - 复用 internal/esindex（不重新实现写入器）；keyset 分页（按主键 id 有序推进，不用 OFFSET）。
+//   - 限速 ≤5k docs/s（默认）；checkpoint 续传（已灌 message_id 高水位，与实时游标物理隔离）。
+//   - 真异常 / ES 永久拒绝落本地 DLQ spill 并精确计数，作为对账门权威输入。
+//   - 配置全走 flag / env。运行结束后建议跑 cmd/reconcile（或 -reconcile 内联对账门）核验。
+//
+// 🔴 隔离纪律：本作业对真实环境执行需 Yu / 运维显式 sign-off 后低峰触发，绝不自动跑生产。
+//
+// 用法：
+//
+//	backfill -mysql-dsn 'user:pass@tcp(host:3306)/im_prod' \
+//	  -tables message,message1,message2,message3,message4 \
+//	  -es http://localhost:9200 -es-index octo-message \
+//	  -spill-dir /var/lib/octo-backfill/dlq -checkpoint /var/lib/octo-backfill/checkpoint.json \
+//	  -rate 5000 -batch 1000 [-reconcile -from <epoch> -to <epoch>]
+package main
+
+import (
+	"context"
+	"database/sql"
+	"flag"
+	"fmt"
+	"log"
+	"os"
+	"os/signal"
+	"strings"
+	"syscall"
+	"time"
+
+	"github.com/Mininglamp-OSS/octo-search-indexer/internal/backfill"
+	"github.com/Mininglamp-OSS/octo-search-indexer/internal/esindex"
+	"github.com/Mininglamp-OSS/octo-search-indexer/internal/recon"
+	_ "github.com/go-sql-driver/mysql"
+	"github.com/opensearch-project/opensearch-go/v3"
+	"github.com/opensearch-project/opensearch-go/v3/opensearchapi"
+)
+
+func main() {
+	log.SetFlags(log.LstdFlags | log.LUTC)
+	if err := run(); err != nil {
+		log.Printf("backfill error: %v", err)
+		os.Exit(1)
+	}
+}
+
+func run() error {
+	var (
+		mysqlDSN   = flag.String("mysql-dsn", os.Getenv("BACKFILL_MYSQL_DSN"), "MySQL DSN (or BACKFILL_MYSQL_DSN)")
+		tablesS    = flag.String("tables", envOr("BACKFILL_TABLES", "message,message1,message2,message3,message4"), "comma-separated message shard tables")
+		esAddrs    = flag.String("es", envOr("BACKFILL_ES", "http://localhost:9200"), "comma-separated OpenSearch addresses")
+		esIndex    = flag.String("es-index", envOr("BACKFILL_ES_INDEX", "octo-message"), "OpenSearch index")
+		esUser     = flag.String("es-user", os.Getenv("BACKFILL_ES_USER"), "OpenSearch username")
+		esPass     = flag.String("es-pass", os.Getenv("BACKFILL_ES_PASS"), "OpenSearch password")
+		spillDir   = flag.String("spill-dir", envOr("BACKFILL_SPILL_DIR", ""), "REQUIRED: local DLQ spill dir for real anomalies / permanent ES rejects")
+		checkpoint = flag.String("checkpoint", envOr("BACKFILL_CHECKPOINT", ""), "checkpoint file path for resumable cursor (empty = in-memory, not resumable)")
+		batch      = flag.Int("batch", envInt("BACKFILL_BATCH", 1000), "keyset batch size")
+		rate       = flag.Float64("rate", envFloat("BACKFILL_RATE", 5000), "ingest rate limit (docs/s; <=0 = unlimited)")
+		backoffMS  = flag.Int("transient-backoff-ms", envInt("BACKFILL_TRANSIENT_BACKOFF_MS", 1000), "ES transient retry backoff base (ms)")
+		timeout    = flag.Duration("timeout", envDuration("BACKFILL_TIMEOUT", 6*time.Hour), "overall timeout")
+		doRecon    = flag.Bool("reconcile", false, "after backfill, run the reconcile gate (requires -from/-to)")
+		fromUnix   = flag.Int64("from", 0, "reconcile window start (epoch seconds, inclusive)")
+		toUnix     = flag.Int64("to", 0, "reconcile window end (epoch seconds, inclusive; 0 = now)")
+	)
+	flag.Parse()
+
+	if *mysqlDSN == "" {
+		return fmt.Errorf("mysql DSN required (-mysql-dsn or BACKFILL_MYSQL_DSN)")
+	}
+	tables := splitCSV(*tablesS)
+	if len(tables) == 0 {
+		return fmt.Errorf("at least one source table required (-tables)")
+	}
+
+	// SIGTERM / SIGINT 优雅停止：当前批结束、checkpoint 已持久化后退出，可续传。
+	ctx, stop := signal.NotifyContext(context.Background(), syscall.SIGINT, syscall.SIGTERM)
+	defer stop()
+	ctx, cancel := context.WithTimeout(ctx, *timeout)
+	defer cancel()
+
+	db, err := sql.Open("mysql", *mysqlDSN)
+	if err != nil {
+		return fmt.Errorf("open mysql: %w", err)
+	}
+	defer func() {
+		if cerr := db.Close(); cerr != nil {
+			log.Printf("warn: close mysql: %v", cerr)
+		}
+	}()
+	if err := db.PingContext(ctx); err != nil {
+		return fmt.Errorf("ping mysql: %w", err)
+	}
+
+	writer, err := esindex.NewWriter(esindex.Config{
+		Addresses: splitCSV(*esAddrs),
+		Index:     *esIndex,
+		Username:  *esUser,
+		Password:  *esPass,
+	})
+	if err != nil {
+		return fmt.Errorf("build ES writer: %w", err)
+	}
+	defer func() {
+		if cerr := writer.Close(); cerr != nil {
+			log.Printf("warn: close ES writer: %v", cerr)
+		}
+	}()
+
+	cp, err := backfill.OpenCheckpoint(*checkpoint)
+	if err != nil {
+		return err
+	}
+	dlq, err := backfill.OpenDLQSpill(*spillDir)
+	if err != nil {
+		return err
+	}
+	defer func() {
+		if cerr := dlq.Close(); cerr != nil {
+			log.Printf("warn: close DLQ spill: %v", cerr)
+		}
+	}()
+
+	runner := backfill.NewRunner(backfill.Config{
+		Tables:           tables,
+		BatchSize:        *batch,
+		DocsPerSec:       *rate,
+		TransientBackoff: time.Duration(*backoffMS) * time.Millisecond,
+	}, backfill.NewMySQLSource(db), writer, cp, dlq)
+
+	log.Printf("backfill start: tables=%v batch=%d rate=%.0f docs/s spill=%s checkpoint=%s index=%s",
+		tables, *batch, *rate, dlq.Path(), *checkpoint, *esIndex)
+
+	stats, runErr := runner.Run(ctx)
+	log.Printf("backfill stats: read=%d indexed=%d (raw_excluded=%d) dlq=%d (payload=%d permanent=%d)",
+		stats.Read, stats.Indexed, stats.RawExcluded, stats.DLQ, stats.DLQPayload, stats.DLQPermanent)
+	if runErr != nil {
+		return fmt.Errorf("backfill run: %w", runErr)
+	}
+
+	if *doRecon {
+		return reconcile(ctx, db, splitCSV(*esAddrs), *esUser, *esPass, *esIndex, tables, *fromUnix, *toUnix, dlq)
+	}
+	log.Printf("backfill complete. Run cmd/reconcile (or re-run with -reconcile -from/-to) to gate correctness. Total DLQ count: %d", dlq.Count())
+	return nil
+}
+
+// reconcile 内联对账门：用 backfill 自己精确统计的 DLQ 计数作为权威输入，避免人工传错 -dlq。
+// 🔴 用**按窗** DLQ 计数（CountInWindow）：reconcile 窗可能不覆盖本次 run 处理的全部行
+// （如分批跑、或只校验某个时间段），用全量 dlqCount 会把窗外的 DLQ 行也减掉 → false
+// mismatch/false OK（codex review P2-window）。
+// 对平退出码 0；不对平返回错误（main 退出码 1）作为 STOP 信号。
+func reconcile(ctx context.Context, db *sql.DB, esAddrs []string, user, pass, index string, tables []string, fromUnix, toUnix int64, dlq *backfill.DLQSpill) error {
+	to := toUnix
+	if to == 0 {
+		to = time.Now().Unix()
+	}
+	if fromUnix > to {
+		return fmt.Errorf("invalid reconcile window: from(%d) > to(%d)", fromUnix, to)
+	}
+	dlqCount := dlq.CountInWindow(fromUnix, to)
+	osClient, err := opensearchapi.NewClient(opensearchapi.Config{
+		Client: opensearch.Config{Addresses: esAddrs, Username: user, Password: pass},
+	})
+	if err != nil {
+		return fmt.Errorf("opensearch client: %w", err)
+	}
+	// 强制 refresh 让刚 bulk 的 doc 对 _count 可见——否则 refresh_interval（默认 1s / backfill
+	// 期建议 -1）未到时 _count 读到旧值，会把对账门误判成漏灌（false MISMATCH）。
+	if _, rerr := osClient.Indices.Refresh(ctx, &opensearchapi.IndicesRefreshReq{Indices: []string{index}}); rerr != nil {
+		return fmt.Errorf("refresh index before reconcile: %w", rerr)
+	}
+	srcCounter := recon.NewMySQLSourceCounter(db, tables)
+	esCounter := recon.NewOSCounter(osClient, index)
+
+	sourceRows, err := srcCounter.CountRows(ctx, fromUnix, to)
+	if err != nil {
+		return err
+	}
+	esDocs, err := esCounter.CountDocs(ctx, fromUnix, to)
+	if err != nil {
+		return err
+	}
+	rawExcluded, err := esCounter.CountRawExcluded(ctx, fromUnix, to)
+	if err != nil {
+		return err
+	}
+	report, err := recon.ReconcileChecked(recon.Counts{
+		SourceRows:  sourceRows,
+		ESDocs:      esDocs,
+		RawExcluded: rawExcluded,
+		DLQ:         dlqCount,
+	})
+	if err != nil {
+		return fmt.Errorf("reconcile inputs not self-consistent: %w", err)
+	}
+	log.Printf("reconcile gate: %s", report.String())
+	if !report.OK {
+		return fmt.Errorf("reconcile MISMATCH (diff=%d): backfill correctness gate FAILED — STOP, do not proceed", report.Diff)
+	}
+	return nil
+}
+
+func envOr(k, def string) string {
+	if v := os.Getenv(k); v != "" {
+		return v
+	}
+	return def
+}
+
+func envInt(k string, def int) int {
+	v := os.Getenv(k)
+	if v == "" {
+		return def
+	}
+	var n int
+	if _, err := fmt.Sscanf(v, "%d", &n); err != nil {
+		log.Printf("backfill: invalid %s=%q, using default %d", k, v, def)
+		return def
+	}
+	return n
+}
+
+func envFloat(k string, def float64) float64 {
+	v := os.Getenv(k)
+	if v == "" {
+		return def
+	}
+	var f float64
+	if _, err := fmt.Sscanf(v, "%g", &f); err != nil {
+		log.Printf("backfill: invalid %s=%q, using default %g", k, v, def)
+		return def
+	}
+	return f
+}
+
+func envDuration(k string, def time.Duration) time.Duration {
+	v := os.Getenv(k)
+	if v == "" {
+		return def
+	}
+	d, err := time.ParseDuration(v)
+	if err != nil {
+		log.Printf("backfill: invalid %s=%q, using default %s", k, v, def)
+		return def
+	}
+	return d
+}
+
+func splitCSV(s string) []string {
+	parts := strings.Split(s, ",")
+	out := make([]string, 0, len(parts))
+	for _, p := range parts {
+		if t := strings.TrimSpace(p); t != "" {
+			out = append(out, t)
+		}
+	}
+	return out
+}
--- a/cmd/reconcile/main.go
+++ b/cmd/reconcile/main.go
@@ -104,12 +104,16 @@ func run() error {
 		return err
 	}
 
-	report := recon.Reconcile(recon.Counts{
+	report, err := recon.ReconcileChecked(recon.Counts{
 		SourceRows:  sourceRows,
 		ESDocs:      esDocs,
 		RawExcluded: rawExcluded,
 		DLQ:         *dlq,
 	})
+	if err != nil {
+		// 输入不自洽（DLQ accounting 加固，阶段 6 (f)）：拒绝在可疑计数上判 OK。
+		return err
+	}
 	fmt.Println(report.String())
 	if !report.OK {
 		// 不对平：退出码 2，作为 backfill/CI gate 的失败信号。
--- a/docs/backfill.md
+++ b/docs/backfill.md
@@ -51,6 +51,13 @@ it lands two classes of "not in ES body index" rows into a **local DLQ spill fil
 That exact count is fed into the reconcile gate as the authoritative `DLQ` input, so
 legitimately-DLQ'd messages are **not** misjudged as ES-missing.
 
+**Crash recovery.** The spill is append-only; each batch fsyncs the spill and then
+atomically advances a durable offset sidecar (`<spill-dir>/backfill-dlq.synced`)
+*before* the checkpoint advances. On restart the spill is truncated to that synced
+offset, discarding the entire un-fsynced dirty suffix (whatever its torn shape) — those
+rows belong to a batch whose checkpoint never advanced, so resume re-derives them. Any
+parse error *within* the synced prefix is treated as real corruption and is fatal.
+
 ## Reconciliation gate (correctness gate)
 
 The job can run the gate inline (`-reconcile -from <epoch> -to <epoch>`) using its own
--- a/docs/backfill.md
+++ b/docs/backfill.md
@@ -1,0 +1,106 @@
+# Historical backfill runbook (`cmd/backfill`)
+
+Phase 6 of the octo message-search pipeline (YUJ-4534). One-shot job that loads the
+historical `message` shard tables into OpenSearch, **bypassing Kafka**, reusing the
+same write path as the live indexer.
+
+```
+message 5 shards → [cmd/backfill: keyset scan → internal/esindex.Writer bulk] → OpenSearch
+```
+
+## Why bypass Kafka
+
+The historical volume (~3.15M rows / ~2.14 GB at ~700 B/row) does not need to go
+through Kafka. Direct bulk-to-ES is faster and avoids any Kafka retention-deletion
+risk during a large one-shot load. The live increment keeps flowing through Kafka
+unchanged; the two paths are safe to run in parallel because the ES doc `_id =
+message_id` makes every write an idempotent upsert — the same message written by both
+paths just overwrites the same doc.
+
+Revoked/deleted messages are **not** a concern for backfill: route 甲 keeps only the
+body in ES and filters revoke/delete at read time via a MySQL join, so even if a
+historically-revoked body is loaded it is filtered on read.
+
+## What it does (correctness properties)
+
+- **Reuses `internal/esindex`** (writer + `EnsureIndex` mapping + `_id=message_id`).
+  No ES write code is re-implemented.
+- **Keyset pagination** by primary-key `id` (`WHERE id>? ORDER BY id ASC LIMIT ?`),
+  never `OFFSET` — O(1) deep-page seeks, no offset drift.
+- **Idempotent** with the live increment (`_id=message_id` upsert; overlap is safe).
+- **Resumable**: a per-shard high-watermark is persisted to a local checkpoint file
+  (atomic temp-file + rename). A watermark is advanced **only after** the batch is
+  terminal (indexed, or routed to the local DLQ spill). The checkpoint is physically
+  isolated from the live increment cursor (`octo_etl_es_cursor`).
+- **Rate-limited** (`-rate`, default ≤5000 docs/s) to avoid overrunning ES / source DB.
+- **payload extraction is byte-for-byte aligned with the live producer** (P1-d three-way
+  split): Signal-encrypted / non-text → `raw_excluded` (still occupies an ES doc, NOT a
+  loss, NOT DLQ); a real parse anomaly → local DLQ spill (counted).
+- **fail-closed**: any DLQ spill-write failure or checkpoint-persist failure STOPs the
+  whole job (real anomalies must never silently vanish — that would corrupt reconcile).
+
+## DLQ accounting (no Kafka DLQ topic on this path)
+
+The live indexer routes poison pills to a Kafka DLQ topic; backfill bypasses Kafka, so
+it lands two classes of "not in ES body index" rows into a **local DLQ spill file**
+(`<spill-dir>/backfill-dlq.ndjson`) and counts them precisely:
+
+1. `payload_unparseable` — a non-encrypted payload that should have parsed but didn't.
+2. permanent ES rejects — per-item 4xx (e.g. mapping conflict) returned by `_bulk`.
+
+That exact count is fed into the reconcile gate as the authoritative `DLQ` input, so
+legitimately-DLQ'd messages are **not** misjudged as ES-missing.
+
+## Reconciliation gate (correctness gate)
+
+The job can run the gate inline (`-reconcile -from <epoch> -to <epoch>`) using its own
+authoritative DLQ count, or you can run `cmd/reconcile` separately. Equation:
+
+```
+ES_docs(window) == source_rows(window) - DLQ(window)
+```
+
+`raw_excluded` docs still occupy an ES doc (content null) and are NOT subtracted;
+revoked/deleted rows keep their ES doc (read-time join) and are NOT subtracted; only
+the DLQ rows (which never reached the ES body index) are subtracted. The gate rejects
+self-inconsistent inputs (e.g. `DLQ > source_rows`, `raw_excluded > es_docs`) rather
+than risk a false OK, and fails (non-zero exit / error) on any mismatch → STOP.
+
+## Suggested ES settings during backfill (octo-deployment)
+
+Per `docs/tuning.md`: `index.refresh_interval=-1`, `index.number_of_replicas=0`
+during the load, then restore and force one `_refresh`, then run the gate.
+
+## Build & run
+
+`cmd/backfill` is an on-demand ops tool (like `cmd/reconcile`); it is **not** baked
+into the long-running `es-indexer` service image. Build it where you run it:
+
+```sh
+go build -o backfill ./cmd/backfill
+
+./backfill \
+  -mysql-dsn 'user:pass@tcp(host:3306)/im_prod' \
+  -tables message,message1,message2,message3,message4 \
+  -es http://opensearch:9200 -es-index octo-message \
+  -spill-dir /var/lib/octo-backfill/dlq \
+  -checkpoint /var/lib/octo-backfill/checkpoint.json \
+  -rate 5000 -batch 1000 \
+  -reconcile -from 1709078400 -to 1718323200
+```
+
+`-spill-dir` is **required** (real anomalies must be durably accounted). Env-var
+equivalents: `BACKFILL_MYSQL_DSN`, `BACKFILL_TABLES`, `BACKFILL_ES`,
+`BACKFILL_ES_INDEX`, `BACKFILL_ES_USER`, `BACKFILL_ES_PASS`, `BACKFILL_SPILL_DIR`,
+`BACKFILL_CHECKPOINT`, `BACKFILL_BATCH`, `BACKFILL_RATE`.
+
+SIGTERM/SIGINT stops cleanly after the current batch (checkpoint persisted) → resumable.
+
+## 🔴 Isolation discipline
+
+Running backfill against a **real** environment requires explicit Yu/ops sign-off and a
+low-peak window. It is never run automatically against production and is decoupled from
+code release — code ships through the normal test→prod path first; backfill is a
+deployment-time ops action verified in the local harness here. The job only **reads**
+the message tables (no schema change, no webhook write path, no binlog); it is
+idempotent, resumable, and pairs with a zero-downtime index alias rebuild.
--- a/harness/README.md
+++ b/harness/README.md
@@ -81,3 +81,37 @@ a real broker + real OpenSearch.
   production the platform provisions them.
 - The harness OpenSearch disables the security plugin for plain-HTTP local
   testing only — production uses authenticated endpoints (`ES_USERNAME`/`ES_PASSWORD`).
+
+## Phase-6 backfill harness (`run-backfill.sh` + `backfill/`)
+
+A second, self-contained e2e that exercises the historical backfill job (`cmd/backfill`)
+against a real MySQL + OpenSearch(IK), bypassing Kafka:
+
+```
+MySQL message shards → cmd/backfill (keyset scan → internal/esindex.Writer bulk) → OpenSearch
+```
+
+| Path | What |
+| --- | --- |
+| `run-backfill.sh` | orchestrator: `up` (throwaway OpenSearch(IK) + MySQL) / `seed` / `backfill` / `verify` / `down` / `all` |
+| `backfill/` | seeds a controlled 6-row suite into the `message` shard tables and verifies the ES result |
+
+The seeded suite is deliberately tiny and explicit so the gate arithmetic is obvious:
+3 text rows (incl. 中文) + 1 Signal-encrypted + 1 non-text (image) + 1 bad-JSON anomaly.
+So `source_rows=6`, expected `ES docs=5` (`6 - 1 DLQ`), `raw_excluded=2`.
+
+```sh
+./harness/run-backfill.sh          # up → seed MySQL → backfill+reconcile gate → verify → down
+KEEP_UP=1 ./harness/run-backfill.sh
+```
+
+What it proves (mirrors the unit tests against real infra):
+- reuses `internal/esindex` writer + IK mapping (`ik_max_word`/`ik_smart`), `_id=message_id`;
+- `raw_excluded` rows (Signal/non-text) still occupy an ES doc; the bad-JSON anomaly is
+  routed to the local DLQ spill and is **positively asserted absent** from ES;
+- the reconcile gate runs inline and is `OK | source_rows=6 es_docs=5 expected=5 diff=0`
+  (it force-refreshes the index first so a refresh lag can't cause a false MISMATCH);
+- re-running is a no-op via the checkpoint and idempotent (`_id` upsert) — ES count stays 5;
+- IK 中文 recall works on backfilled docs (`公园` → the 公园 row, `北京` → the 北京 row).
+
+See `docs/backfill.md` for the production runbook and isolation discipline.
--- a/harness/backfill/main.go
+++ b/harness/backfill/main.go
@@ -1,0 +1,279 @@
+// Command backfill-harness seeds a controlled set of message rows into a local MySQL
+// (mimicking the octo-im message shard schema) for the phase-6 backfill e2e, then
+// (optionally) verifies the result in OpenSearch after cmd/backfill has run.
+//
+// Isolation: connects only to the local throwaway MySQL + OpenSearch. Never a shared env.
+//
+// Modes:
+//
+//	-mode seed    create shard tables + insert a controlled suite, print the source row count
+//	-mode verify  assert ES doc count / raw_excluded / DLQ-bypassed match the seeded suite
+package main
+
+import (
+	"context"
+	"database/sql"
+	"encoding/json"
+	"flag"
+	"fmt"
+	"log"
+	"os"
+	"strings"
+	"time"
+
+	_ "github.com/go-sql-driver/mysql"
+	"github.com/opensearch-project/opensearch-go/v3"
+	"github.com/opensearch-project/opensearch-go/v3/opensearchapi"
+)
+
+// Seeded suite shape (kept tiny + explicit so the gate arithmetic is obvious):
+//   - 3 normal text rows (incl. 中文)            → indexed, content present
+//   - 1 Signal-encrypted row (signal column)     → raw_excluded, still 1 ES doc
+//   - 1 non-text (image) row                     → raw_excluded, still 1 ES doc
+//   - 1 bad-JSON non-encrypted row (real anomaly)→ DLQ spill, NOT in ES
+//
+// So: source_rows=6, expected ES docs=5 (6 - 1 DLQ), raw_excluded=2.
+const (
+	wantSource      = 6
+	wantESDocs      = 5
+	wantRawExcluded = 2
+	wantDLQ         = 1
+)
+
+func main() {
+	log.SetFlags(log.LstdFlags | log.LUTC)
+	var (
+		mode   = flag.String("mode", "seed", "seed|verify")
+		dsn    = flag.String("mysql-dsn", envOr("BF_MYSQL_DSN", "root:root@tcp(localhost:23307)/im_test"), "MySQL DSN")
+		esURL  = flag.String("es", envOr("BF_ES", "http://localhost:29200"), "OpenSearch URL")
+		index  = flag.String("es-index", envOr("BF_ES_INDEX", "octo-message"), "index")
+		tables = flag.String("tables", envOr("BF_TABLES", "message,message1"), "shard tables")
+		base   = flag.Int64("base-ts", time.Now().Unix(), "base created_at epoch seconds")
+	)
+	flag.Parse()
+	if err := run(*mode, *dsn, *esURL, *index, splitCSV(*tables), *base); err != nil {
+		log.Fatalf("backfill-harness %s: %v", *mode, err)
+	}
+}
+
+func run(mode, dsn, esURL, index string, tables []string, base int64) error {
+	ctx, cancel := context.WithTimeout(context.Background(), 60*time.Second)
+	defer cancel()
+	switch mode {
+	case "seed":
+		return seed(ctx, dsn, tables, base)
+	case "verify":
+		return verify(ctx, esURL, index, base)
+	default:
+		return fmt.Errorf("unknown mode %q", mode)
+	}
+}
+
+func seed(ctx context.Context, dsn string, tables []string, base int64) error {
+	db, err := sql.Open("mysql", dsn)
+	if err != nil {
+		return err
+	}
+	defer mustClose(db)
+	if err := db.PingContext(ctx); err != nil {
+		return err
+	}
+	if len(tables) == 0 {
+		return fmt.Errorf("need at least one table")
+	}
+	for _, t := range tables {
+		if err := createTable(ctx, db, t); err != nil {
+			return err
+		}
+	}
+	// Put the 6-row suite across the shard tables (round-robin-ish; deterministic).
+	t0 := tables[0]
+	t1 := tables[len(tables)-1]
+	rows := []struct {
+		table       string
+		mid         string
+		fromUID     string
+		channelID   string
+		channelType uint8
+		setting     uint8
+		signal      int
+		payload     string
+	}{
+		{t0, "h-en", "u1", "g1", 2, 0, 0, `{"type":1,"content":"hello world pipeline"}`},
+		{t0, "h-zh1", "u1", "g1", 2, 0, 0, `{"type":1,"content":"今天天气很好我们去公园散步吧"}`},
+		{t1, "h-zh2", "u2", "g1", 2, 0, 0, `{"type":1,"content":"搜索引擎中文分词测试北京欢迎你"}`},
+		{t1, "h-signal", "u3", "u3@u4", 1, 0, 1, `ENCRYPTED-NOT-JSON`},
+		{t0, "h-image", "u1", "g1", 2, 0, 0, `{"type":2,"url":"http://x/y.png"}`},
+		{t1, "h-bad", "u2", "g1", 2, 0, 0, `{not valid json`},
+	}
+	for i, r := range rows {
+		if _, err := db.ExecContext(ctx, fmt.Sprintf(
+			"INSERT INTO `%s` (message_id, from_uid, channel_id, channel_type, setting, `signal`, `timestamp`, created_at, payload) "+
+				"VALUES (?,?,?,?,?,?,?,FROM_UNIXTIME(?),?)", r.table),
+			r.mid, r.fromUID, r.channelID, r.channelType, r.setting, r.signal, base, base+int64(i), r.payload,
+		); err != nil {
+			return fmt.Errorf("insert %s: %w", r.mid, err)
+		}
+	}
+	log.Printf("seeded %d source rows across %v (expected ES docs=%d raw_excluded=%d DLQ=%d)",
+		wantSource, tables, wantESDocs, wantRawExcluded, wantDLQ)
+	return nil
+}
+
+func createTable(ctx context.Context, db *sql.DB, table string) error {
+	// 表名来自 -tables flag（本地 harness 控制），但 DROP/CREATE 仍白名单校验防注入。
+	if !safeTableName(table) {
+		return fmt.Errorf("unsafe table name %q", table)
+	}
+	// Minimal shape mirroring the columns cmd/backfill reads.
+	_, err := db.ExecContext(ctx, fmt.Sprintf("DROP TABLE IF EXISTS `%s`", table))
+	if err != nil {
+		return err
+	}
+	_, err = db.ExecContext(ctx, fmt.Sprintf(
+		"CREATE TABLE `%s` ("+
+			"id BIGINT AUTO_INCREMENT PRIMARY KEY, "+
+			"message_id VARCHAR(20) NOT NULL, "+
+			"from_uid VARCHAR(40) NOT NULL DEFAULT '', "+
+			"channel_id VARCHAR(100) NOT NULL DEFAULT '', "+
+			"channel_type TINYINT UNSIGNED NOT NULL DEFAULT 0, "+
+			"setting TINYINT UNSIGNED NOT NULL DEFAULT 0, "+
+			"`signal` INT NOT NULL DEFAULT 0, "+
+			"`timestamp` BIGINT NOT NULL DEFAULT 0, "+
+			"created_at DATETIME NOT NULL, "+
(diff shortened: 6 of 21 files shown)


Jerry-Xin (approved):
The PR is in scope and the implementation is generally consistent with the existing indexer architecture: it reuses `internal/esindex`, uses keyset pagination, and adds focused tests for extraction, DLQ replay/deduping, checkpointing, retry behavior, and reconciliation.

💬 Non-blocking
- 🟡 Warning: [internal/recon/recon.go:8](/Users/mlamp/.openclaw/tmp/pr-review-akclhvrt/repo/internal/recon/recon.go:8) still documents the invariant as `ES + DLQ + raw_excluded == source`, but the implemented and intended equation is `ES == source - DLQ` because `raw_excluded` documents are included in ES. The code is right; the package comment should be corrected to avoid future operator confusion.
- 🟡 Warning: [cmd/backfill/main.go:68](/Users/mlamp/.openclaw/tmp/pr-review-akclhvrt/repo/cmd/backfill/main.go:68) says inline reconcile “requires -from/-to”, but the command accepts `-reconcile` with the default `from=0`. That can accidentally run a whole-history gate and include rows outside the intended backfill window. Consider rejecting `-reconcile` unless `-from` is explicitly set.
- 🔵 Suggestion: [internal/backfill/checkpoint.go:143](/Users/mlamp/.openclaw/tmp/pr-review-akclhvrt/repo/internal/backfill/checkpoint.go:143) renames the checkpoint atomically but does not fsync the parent directory after rename. Because ES writes and DLQ writes are idempotent/deduped, the likely failure mode is redundant reprocessing rather than data loss, so this is not blocking.
- 🔵 Suggestion: [internal/backfill/runner.go:194](/Users/mlamp/.openclaw/tmp/pr-review-akclhvrt/repo/internal/backfill/runner.go:194) permanent ES reject DLQ records lose the source row ID and original payload because `writeMain` only has `searchmsg.Message`. Accounting still works for normal `message_id` values, but preserving source metadata would make replay and triage stronger.

✅ Highlights
- The PR belongs in this repository: it extends the message indexing/reconciliation tooling for `octo-search-indexer`.
- Strong reuse of the existing ES writer avoids a second bulk-write implementation.
- DLQ spill replay, deduping, windowed counts, and fsync-before-checkpoint are well covered.
- `ReconcileChecked` fails closed for impossible counts, which is the right direction for an ops gate.

Verification run: `go test ./...` passed.

lml2468:
## ✅ Review — octo-search-indexer PR #8 (9627d8c): historical backfill + DLQ-accounting reconcile gate — strong, good to merge

Cloned, built, go test -race ./internal/backfill ./internal/recon green. Large but well-structured (every backfill submodule has tests). Hard correctness concerns all handled. No blocking findings.

### Verified — the parts that matter
- Reconcile gate rejects self-inconsistent inputs (recon.go Counts.Validate): fails on DLQ>SourceRows ("a too-high DLQ would shrink Expected and mask a shortfall as false OK"), RawExcluded>ESDocs, negatives, SourceRows-DLQ<0. OK := diff==0 (strict). The right defense against the gate false-passing when DLQ accounting is wrong — refuses contradictory inputs instead of computing a misleading OK.
- DLQ spill durability, fail-closed (backfill/dlq.go): any spill write failure returns error → caller STOPs; O_APPEND, content fsync via Sync/Close, directory-entry fsync on creation, replay-on-reopen deduped by message_id (empty id → table:id).
- Idempotent + resumable: ES _id=message_id makes every write an upsert → safe parallel with the live Kafka increment; per-shard checkpoint (atomic, isolated from live cursor) → re-runs no-ops.
- SQL safety: every dynamic table name (source SELECT + harness DROP/CREATE) safeTableName-allowlisted ([A-Za-z0-9_]) + backtick-wrapped; time/paging via placeholders. Keyset pagination (WHERE id>? ORDER BY id ASC LIMIT ?, no OFFSET).
- Scope: cmd/backfill + harness not baked into the es-indexer image (precedent: cmd/reconcile); production Dockerfile unchanged; job only reads message tables.

### Non-blocking
- Shared recon UNIX_TIMESTAMP non-sargable + empty-tables fail-closed are pre-existing #7 follow-ups.
- Jerry-Xin's 🟡s (pkg-comment ES/DLQ wording; -reconcile should require -from) are fair doc/arg-validation nits.

LGTM.

lml2468 (approved):
Verified: reconcile gate Counts.Validate rejects self-inconsistent inputs (DLQ>source, raw_excluded>es) so DLQ accounting can't false-pass a shortfall; DLQ spill fail-closed with content+dir fsync and message_id-deduped replay; _id=message_id upsert idempotent + parallel-safe with live increment; per-shard checkpoint resumable; all dynamic table names allowlisted; keyset pagination no OFFSET. build + go test -race green. LGTM.

mochashanyao (approved):
[Octo-Q · automated review]

---

# octo-search-indexer PR#8 Review Report
**PR:** https://github.com/Mininglamp-OSS/octo-search-indexer/pull/8
**Head SHA:** `9627d8cbbeb2155d1d5d55899e9b5ed566120673`
**Title:** feat(backfill): historical message backfill job + DLQ-accounting reconcile gate
**Routing:** complexity=complex (automated review)

---

## 1. Verification Summary

| Item | Status | Evidence |
|------|--------|----------|
| Keyset pagination (no OFFSET) | ✅ | `internal/backfill/source.go:47` — `WHERE id>? ORDER BY id ASC LIMIT ?` |
| Table name injection prevention | ✅ | `safeTableName` allowlist `[A-Za-z0-9_]` in `source.go:72`, `harness/backfill/main.go:268` |
| ES idempotent upsert (`_id=message_id`) | ✅ | Reuses `internal/esindex.Writer` (existing `writer.go:40-48`) |
| Payload extraction aligned with producer | ✅ | `extract.go:38-44` constants + `extract_test.go:47-52` locks alignment |
| DLQ dedup by message_id | ✅ | `dlq.go:155-160` — `seen[key]` check before append |
| DLQ replay on reopen | ✅ | `dlq.go:93-120` — `loadSeen` scans existing file |
| Durability ordering (DLQ fsync → checkpoint advance) | ✅ | `runner.go:131-138` — `dlq.Sync()` before `cp.Advance()` |
| Fail-closed on DLQ write failure | ✅ | `dlq.go:163-168` — returns error; `runner.go:155-158` propagates |
| Reconcile gate input validation | ✅ | `recon.go:52-68` — `Validate()` rejects impossible inputs |
| ReconcileChecked adoption in cmd/reconcile | ✅ | `cmd/reconcile/main.go:107-114` — error propagated |
| Inline reconcile force-refreshes ES | ✅ | `cmd/backfill/main.go:175-178` — `Indices.Refresh` before count |
| Windowed DLQ count for reconcile | ✅ | `cmd/backfill/main.go:163` — `dlq.CountInWindow(fromUnix, to)` |
| Checkpoint monotonic advance | ✅ | `checkpoint.go:82-85` — `newID <= current` silently ignored |
| Checkpoint atomic write (tmp+rename) | ✅ | `checkpoint.go:113-147` |
| Rate limiter token bucket | ✅ | `ratelimit.go:47-85` — correct with fake-clock tests |
| Transient retry in-place (subset only) | ✅ | `runner.go:183-213` — `next` collects only unresolved items |
| Permanent ES reject → DLQ | ✅ | `runner.go:196-210` — `res.Permanent()` → spill |
| Signal detection (column + setting bit) | ✅ | `extract.go:131-135` — matches producer dual-check |
| raw_excluded still occupies ES doc | ✅ | `runner.go:159-163` — appended to `main` batch |
| Test coverage | ✅ | 474-line runner_test + 183-line dlq_test + 144-line extract_test + 70-line checkpoint_test + 48-line ratelimit_test + 50-line recon_test |

## 2. Findings

### F1 — P2: Checkpoint persist missing directory fsync after rename

**File:** `internal/backfill/checkpoint.go:113-147` (`persistLocked`)

The checkpoint atomic write does `CreateTemp → Write → Sync → Close → Rename` but does **not** fsync the parent directory after the rename. Compare with `dlq.go:87-95` which explicitly calls `fsyncDir(dir)` after creating the spill file.

**Impact:** On filesystems without guaranteed rename atomicity (XFS, ext4 with `data=writeback`), a host crash immediately after rename could lose the new checkpoint, reverting to the old position. Due to ES `_id` idempotent upsert + DLQ dedup, reprocessing produces correct results — just redundant work.

**Recommendation:** Add `fsyncDir(filepath.Dir(s.path))` after the rename in `persistLocked`, reusing the same `fsyncDir` pattern from `dlq.go`. Low-risk hardening.

### F2 — P2: DLQ record for permanent ES reject omits source table auto-increment ID

**File:** `internal/backfill/runner.go:198-205`

```go
rec := dlqRecord{
    Reason: "permanent_es_reject",
    Table:  table, MessageID: remaining[i].MessageID,
    CreatedAt: remaining[i].CreatedAt,
}
```

The `ID` field (source table auto-increment PK) is not set (defaults to 0). The `dedupKey` function (`dlq.go:125-130`) falls back to `fmt.Sprintf("%s:%d", rec.Table, rec.ID)` when `MessageID` is empty. If a permanent reject has empty `MessageID`, its dedup key becomes `"table:0"`, potentially colliding with other permanent rejects that also have empty `MessageID`.

**Impact:** Extremely unlikely in practice — `message_id` is a NOT NULL column in the message schema. But if it ever occurs, distinct permanent rejects could be incorrectly deduplicated, causing DLQ undercount.

**Recommendation:** Populate `ID` from the original source row. This requires threading the source row's `ID` through to `writeMain` (e.g., a parallel `[]int64` of source IDs, or a map from `MessageID` to source `ID`).

### F3 — P2 (documented limitation): Validate() cannot detect artificially-inflated DLQ masking a shortfall

**File:** `internal/recon/recon_test.go:118-131` (`TestReconcileChecked_DLQMasksShortfall`)

The test explicitly documents that if DLQ=100 ≤ SourceRows=1000 and ESDocs=900, `Validate()` passes and `Reconcile` returns false OK (Expected=900==ESDocs=900). The defense is that the DLQ count is authoritative (counted by the backfill itself, not human-supplied). This is a known limitation, well-documented in the test comments.

**No action required** — the authoritative DLQ counting by the backfill job itself is the correct defense.

## 3. Suggestions

1. **F1 fix:** Add directory fsync after checkpoint rename for parity with DLQ spill durability.
2. **F2 fix:** Thread source row ID into `writeMain` for permanent reject DLQ records.
3. **Minor:** Consider extracting `safeTableName` to a shared internal package to avoid drift across `internal/backfill`, `internal/recon`, and `harness/backfill`.

## 4. Additional Observations

- **Live increment isolation:** Backfill uses its own local checkpoint file, completely isolated from the live `octo_etl_es_cursor`. Both paths write ES with `_id=message_id` (idempotent upsert). Parallel execution is safe. ✅
- **Read-only on source:** Backfill only issues `SELECT` queries against message tables. No schema changes, no writes. ✅
- **Harness quality:** The e2e harness (`harness/run-backfill.sh` + `harness/backfill/main.go`) seeds a controlled 6-row suite with explicit arithmetic (6 source → 5 ES docs → 2 raw_excluded → 1 DLQ) and positively asserts the bad-JSON row is absent from ES. Strong verification.
- **Test `TestRunner_ResumeAfterDLQNoDoubleCount`:** Specifically tests the crash-between-DLQ-write-and-checkpoint-advance scenario with empty checkpoint + same spill dir. Validates dedup works correctly. ✅

## 5. Data-Flow Trace

|---|---|---|
| `srcMessageRow.Payload` | `MySQLSource.ReadBatch` → `rows.Scan(&r.Payload)` (source.go:60) | ✅ — raw bytes from MySQL BLOB column |
| `srcMessageRow.Signal` | `MySQLSource.ReadBatch` → `rows.Scan(&r.Signal)` (source.go:59) | ✅ — INT column |
| `srcMessageRow.Setting` | `MySQLSource.ReadBatch` → `rows.Scan(&r.Setting)` (source.go:59) | ✅ — TINYINT column |
| `srcMessageRow.CreatedUnix` | `UNIX_TIMESTAMP(created_at)` in SQL (source.go:48-49) | ✅ — epoch seconds |
| `extractMessage` outcome | `isSignalEncrypted(row)` checks Signal + Setting bit5 (extract.go:131-135) | ✅ — matches producer dual-check |
| `contentTypeText` constant | Hardcoded `= 1` (extract.go:39), test-pinned to octo-lib `common.Text` (extract_test.go:47-49) | ✅ |
| `signalSettingMask` constant | Hardcoded `= 1<<5` (extract.go:44), test-pinned (extract_test.go:50-52) | ✅ |
| DLQ `CountInWindow` | Iterates `seen` map (message_id → created_at), filters `createdAt >= from && <= to` (dlq.go:189-198) | ✅ — inclusive bounds match ES `gte/lte` and MySQL `BETWEEN` |
| Reconcile `SourceRows` | `recon.MySQLSourceCounter.CountRows` → `COUNT(*) WHERE UNIX_TIMESTAMP(created_at) BETWEEN ? AND ?` per table (collectors.go:52-62) | ✅ |
| Reconcile `ESDocs` | `recon.OSCounter.CountDocs` → ES `_count` with `range` filter on `created_at` (collectors.go:97-99) | ✅ |
| Reconcile `DLQ` | `dlq.CountInWindow(fromUnix, to)` — authoritative count from spill file | ✅ |
| Checkpoint `LastID` | Advanced to `rows[len(rows)-1].ID` after DLQ sync (runner.go:137-138) | ✅ — monotonic, atomic persist |
| `BulkItemResult.Permanent()` | `isPermanentStatus`: 4xx except 429 (writer.go:72-79) | ✅ — existing code, reused correctly |

## 6. R5 Blind-Spot Checklist

- **C1 — Dual-path parity:** N/A. This PR adds a new one-shot ops tool, not a symmetric add/remove path. The backfill path is explicitly isolated from the live increment (separate cursor, idempotent ES writes).
- **C2 — Control-flow ordering / nesting reuse:** Clear. The `processBatch → writeMain` flow is linear. DLQ Sync → checkpoint Advance ordering is correct and tested. No nested reuse of safety functions.
- **C3 — Authorization boundary ≠ capability boundary:** N/A. This is a standalone CLI tool requiring explicit MySQL DSN + ES credentials + ops sign-off. No new endpoints or exposed capabilities.

---

**[Octo-Q] verdict: APPROVE**

No P0/P1 findings. Three P2 observations (checkpoint dir fsync, DLQ record ID omission, documented Validate limitation) are robustness hardening, not correctness blockers. The PR demonstrates strong engineering discipline: idempotent design, fail-closed semantics, durability ordering, comprehensive test coverage, and well-documented limitations.

lml2468:
## Correction — mochashanyao's checkpoint dir-fsync 🔴 is real; I missed it.

mochashanyao is right; I should not have approved without checking this. Verified:

### 🔴 Checkpoint rename lacks the parent-dir fsync that DLQ has
checkpoint.go:130-146 does the right durable-write prefix — CreateTemp → tmp.Sync() (content fsync) → tmp.Close() → os.Rename(tmp, path) (atomic) — but returns with NO fsyncDir(dir) after the rename. So the directory-entry change from the rename can be lost on a host crash: file content is durable, but the rename pointing the checkpoint name at it may not be.

Irony: the DLQ module already does this right — dlq.go:74 calls fsyncDir(dir) after creating the spill file, and there's a reusable fsyncDir helper (dlq.go:126). Checkpoint just didn't apply the same step to its rename. mochashanyao's 🟡#3 ("unify DLQ and checkpoint fsync") is the precise framing — fix is one line: fsyncDir(dir) after the rename, reusing the existing helper.

### Severity — should-fix, NOT a data-integrity blocker
Checkpoint is only for resumability. A lost checkpoint → job resumes from an older point and re-does part of the backfill — safe because backfill is idempotent (_id=message_id upsert). Consequence is redundant work, not data loss. mochashanyao's 🔴 is correct on durability grounds; high-value should-fix (one line, reusing existing code), given the matching DLQ path already sets the precedent.

### Other notes
- DLQ record lacks source-table auto-increment id (I'd say 🟡): records dedup by message_id (= ES _id), the correct dedup key; adding table:id aids traceability but isn't a correctness gap.
- Unify fsync strategy (🟡): same as the checkpoint fix.

### Correcting myself
My approve verified the DLQ spill fsync (correct) but didn't separately check the checkpoint rename durability — same class of miss as assuming a sibling path is fine because the adjacent one is. Add the checkpoint dir-fsync; trivial fix, fsyncDir helper is right there. Data is safe regardless due to idempotency, so fast-follow is acceptable, but worth folding in.

yujiawei (commented):
# Code Review — PR #8 (octo-search-indexer)

Reviewed at head `9627d8c` against `main` (merge-base `c7b4b45`). `go build`, `go vet`, and `go test ./internal/backfill/... ./internal/recon/...` are all green locally.

This is a well-engineered backfill job. The design discipline is real: it reuses `internal/esindex` rather than reimplementing the writer, uses keyset pagination (no `OFFSET`), the `_id=message_id` upsert makes it safe to run alongside the live Kafka indexer, the reconcile gate is hardened against self-inconsistent inputs, and the durability ordering (DLQ `Sync()` before checkpoint `Advance()`) is correct and deliberate. Payload extraction is pinned to the live producer's vectors via test, which is the right way to keep the two paths byte-aligned.

The one blocker is on the crash-recovery side of that same DLQ subsystem: the replay path is fatal-on-corrupt, which defeats the automated-resume guarantee the rest of the design works hard to provide.

## Blocking

### [P1] DLQ replay wedges the job after a mid-batch crash — `internal/backfill/dlq.go:107`

`loadSeen` treats any `json.Unmarshal` error on a spill line as fatal and fails the whole `OpenDLQSpill`:

```go
if err := json.Unmarshal(line, &rec); err != nil {
    return nil, fmt.Errorf("backfill: corrupt spill line during replay: %w", err)
}
```

But `Write` (dlq.go:158) appends records with a plain `s.f.Write(...)` and the fsync only happens once per batch in `Sync()` (called at runner.go:120, *before* `checkpoint.Advance` at runner.go:125). So the documented failure mode — process/host crash mid-batch, after some `Write`s but before `Sync()` — can leave a torn final NDJSON line on disk.

On restart, `OpenDLQSpill` hits that partial line and refuses to start. This is exactly the scenario the durability ordering was built to recover from: because the checkpoint has *not* advanced past those rows, they will be re-read and re-written safely (idempotent via `_id=message_id` and the dedup set). The job *should* auto-resume, but instead it wedges and needs an operator to manually truncate the last line. For a multi-hour, 3.15M-row job (default `-timeout 6h`), a mid-run crash is the expected failure mode, not an exotic one — and the PR explicitly advertises "resumable per-shard checkpoint."

Fix: during replay only, tolerate a single unparseable *final* line (truncate/ignore the unsynced tail) instead of returning a fatal error. A torn line can only be the last line (appends are sequential and prior lines were fsynced before the checkpoint advanced past them), so dropping an unparseable tail is safe — the corresponding rows sit above the checkpoint and will be reprocessed. Keep the error fatal for a corrupt line in the *interior* of the file (that signals real corruption, not a torn tail). Worth a test that appends a half-written line and asserts `OpenDLQSpill` recovers.

## Non-blocking

### [P2] Permanent-reject DLQ records lose source PK and original payload — `internal/backfill/runner.go:194-201`

When `_bulk` returns a permanent 4xx, the DLQ record is built from the already-transformed `searchmsg.Message`, after the `*srcMessageRow` is out of scope:

```go
rec := dlqRecord{
    Reason: "permanent_es_reject",
    Table:  table, MessageID: remaining[i].MessageID,
    CreatedAt: remaining[i].CreatedAt,
}
if remaining[i].Content != nil {
    rec.Payload = []byte(*remaining[i].Content)
}
```

Two consequences: (1) `ID` is always 0 and `Payload` is the normalized content (or nil for raw-excluded docs), so the spill record no longer carries the source table PK or the original payload needed for triage/replay — the whole point of spilling. (2) If a rejected doc has an empty `MessageID`, `dedupKey` falls back to `table:id` = `table:0`, so *all* empty-id permanent rejects collapse into one dedup key and undercount. Counting stays correct for the common case (non-empty message_id), so this is advisory, but it weakens the recovery value of the permanent-reject branch. Consider threading the originating `*srcMessageRow` (or a slice aligned to `main`) through `writeMain` so the permanent-reject path can spill the same `{id, payload}` the payload-unparseable path does.

### [P2] Checkpoint crashes late if the parent dir doesn't exist — `internal/backfill/checkpoint.go:113-114`

`OpenCheckpoint` happily accepts a path whose parent directory is missing (it only `ReadFile`s, tolerating `IsNotExist`). But `persistLocked` does `os.CreateTemp(filepath.Dir(s.path), ...)`, which fails if that dir doesn't exist. The result: the job starts, reads MySQL, writes the first batch to ES, and only then crashes on the first `Advance()` — wasting work and failing late. (The DLQ spill dir is handled correctly — `OpenDLQSpill` does `MkdirAll`.) Mirror that: `MkdirAll(filepath.Dir(path))` in `OpenCheckpoint`, or at least stat the parent up front and fail fast.

## Coverage notes

- Verified the rate limiter is **not** a deadlock or data-race risk: `burst` is wired to `BatchSize` (runner.go:70), and `Wait` is only ever called with `len(rows) ≤ BatchSize` (the source `LIMIT ?` bounds it), so `need ≤ burst` always holds; and the table loop (runner.go:84) is sequential with a single `Wait` call site, so there's no concurrent access to the token-bucket state. No change needed there.
- `safeTableName` allowlisting is applied on every dynamic table name (source reads and the recon counter), so the `fmt.Sprintf` of table names into SQL is not injectable from this path; table names come from operator config, not user input.
- Not independently exercised: `harness/run-backfill.sh` end-to-end (needs Docker MySQL + OpenSearch/IK) and the real-environment behavior of `internal/esindex.Writer.Bulk` permanent/transient classification under live ES. The PR body reports a green e2e run; I reviewed the harness code but did not run it.

## Verdict

Changes requested for the [P1] replay-wedge — it's a small, well-bounded fix and it closes the one gap between this PR's stated crash-safety guarantee and its actual behavior. The two P2s are worth folding into the same revision but are not merge blockers on their own.

lml2468:
## #8 consolidated — verified all open claims against the code (one I'd earlier mis-accepted)

Checked each against head 9627d8c:

### 🔴 REAL — checkpoint rename lacks dir fsync (mochashanyao #1)
persistLocked does tmp.Sync()→Close()→Rename with no fsyncDir after the rename, while DLQ's writer (dlq.go:74,126) does exactly that and comments why. Real crash-durability gap, this-PR-introduced. Softened (not eliminated) by idempotency: lost checkpoint → re-process completed rows → no dupes/loss via _id=message_id upsert. One-line fix reusing existing fsyncDir. Steve and I both missed it (verified DLQ's fsync, didn't check the sibling checkpoint path).

### 🔴 REAL — replay fails hard on a corrupt spill line (yujiawei, new)
dlq.go:107-109: during replay, a line that fails json.Unmarshal → return nil,err → reopen fails → backfill can't start. The irony: the spill exists FOR crash recovery, but a single half-written line (kill -9/disk-full mid-write that fsync can't fully exclude) makes recovery impossible until someone hand-edits the file. A recovery mechanism that won't recover from its own partial write is self-defeating. Fix: skip + log/alert the corrupt line and continue replay (don't fail-hard). Valid should-fix.

### ❌ FALSE-AS-STATED — "DLQ record missing source auto-inc ID / payload" (mochashanyao #2 + yujiawei 🟡)
dlqRecord (dlq.go:18-26) already has ID int64 // 源自增 id, plus Table, MessageID, and Payload []byte // 原始 payload（供排查/回灌）. Source PK AND raw payload are both already persisted in every DLQ record. So "missing source PK / original payload" is incorrect — the fields are there. Correcting myself: I reflexively accepted mochashanyao #2 in my prior reply without opening the struct — that was wrong; Steve checked the struct and is right it's false-as-stated. Dropping #2.

### Net
Two real should-fixes — checkpoint dir-fsync (durability, idempotency-softened) and replay-skip-corrupt-line (recovery robustness); both small. The "missing source id/payload" claims are false (fields exist). Neither real issue is data-corrupting (idempotent writes + fail-closed spill), but both undermine the crash-recovery story this PR is built around, so worth fixing before merge. Recommend: (1) fsyncDir after checkpoint rename, (2) replay skips+logs corrupt lines instead of aborting.

mochashanyao (approved):
[Octo-Q · automated review]

---

# octo-search-indexer PR#8 Review Report
**PR:** Mininglamp-OSS/octo-search-indexer#8
**Head SHA:** `72ee8fb66d5002a43041062ad09410571346b7bd`
**Title:** feat(backfill): historical message backfill job + DLQ-accounting reconcile gate
**Scope:** 21 files, +3181/-6 lines (new `cmd/backfill`, `internal/backfill/*`, hardened `internal/recon`, harness)

---

## 1. Verification Summary

| Area | Status | Evidence |
|------|--------|----------|
| Keyset pagination (no OFFSET) | ✅ | `source.go:47` — `WHERE id>? ORDER BY id ASC LIMIT ?` |
| ES write reuse (`internal/esindex`) | ✅ | `runner.go:91` — `r.writer.Bulk(ctx, msgs)`; no ES write code reimplemented |
| Idempotent upsert (`_id=message_id`) | ✅ | `esindex/writer.go` uses `index` action (not `create`); `runner_test.go:TestRunner_Idempotent` verifies |
| Payload extraction alignment with producer | ✅ | `extract.go` 3-way split matches producer; `extract_test.go:TestExtract_ConstantsMatchOctoLib` pins `contentTypeText==1`, `signalSettingMask==32` |
| DLQ dedup by message_id | ✅ | `dlq.go:Write()` checks `seen[key]` before append; `dlq_test.go:TestDLQSpill_IdempotentByMessageID` |
| DLQ replay on reopen | ✅ | `dlq.go:recoverAndLoad()` truncates to synced offset, replays clean prefix; `dlq_test.go:TestDLQSpill_ReplayCountOnReopen` |
| Windowed DLQ count for reconcile | ✅ | `dlq.go:CountInWindow()` iterates `seen` values by `created_at`; aligns with recon `rangeFilter` gte/lte |
| Durability ordering (DLQ fsync → checkpoint advance) | ✅ | `runner.go:runTable()` — `processBatch` → `dlq.Sync()` → `cp.Advance()` |
| Crash recovery (dirty suffix truncation) | ✅ | `dlq.go:recoverAndLoad()` truncates to `syncedLen`; `dlq_test.go:TestDLQSpill_RecoverDirtySuffix` |
| Reconcile gate hardening | ✅ | `recon.go:ReconcileChecked()` validates self-consistency before arithmetic; `recon_test.go` covers DLQ>source, raw_excluded>ES, negative |
| Fail-closed on DLQ spill failure | ✅ | `dlq.go:Write()` returns error on write failure; `runner.go:processBatch()` propagates → STOP |
| Table name injection prevention | ✅ | `safeTableName()` in `source.go:78`, `harness/backfill/main.go:271` — allowlist `[A-Za-z0-9_]` |
| Signal detection (dual source) | ✅ | `extract.go:isSignalEncrypted()` checks both `row.Signal != 0` and `row.Setting & (1<<5)` |
| Transient retry (in-place, subset) | ✅ | `runner.go:writeMain()` — only unresolved items retry; `runner_test.go:TestRunner_TransientRetriesInPlace` |
| Permanent ES reject → DLQ | ✅ | `runner.go:writeMain()` — `res.Permanent()` → `dlq.Write()` with source PK + payload |

## 2. Findings

### No P0 or P1 findings.

### P2 Findings

**P2-1: Checkpoint `persistLocked` omits directory fsync after rename**
- **File:** `internal/backfill/checkpoint.go:148-153`
- **What:** After `os.Rename(tmpName, s.path)`, the parent directory is not fsync'd. Compare with `dlq.go:writeSyncedOffset()` which calls `fsyncDir(dir)` after rename.
- **Risk:** On some filesystems, if the host crashes immediately after rename, the directory entry might not be persisted. The checkpoint file could be "lost" (still on disk but not visible in the directory).
- **Impact:** Low — losing the checkpoint only means re-processing from the old watermark. ES writes are idempotent (`_id=message_id` upsert), so re-processing is safe. This is an efficiency concern, not a correctness concern.
- **Diff-scope:** New code in this PR.
- **Suggestion:** Add `fsyncDir(dir)` after the rename for consistency with the DLQ sidecar path.

**P2-2: `safeTableName` duplicated across 3 files**
- **Files:** `internal/backfill/source.go:78`, `internal/recon/collectors.go:68`, `harness/backfill/main.go:271`
- **What:** Three identical copies of the same allowlist function.
- **Impact:** Maintainability — if the allowlist ever changes, all three must be updated. Low risk given the simplicity and stability of this function.
- **Diff-scope:** New duplication in this PR (backfill + harness copies; recon copy pre-existed).
- **Suggestion:** Consider extracting to a shared internal package if more consumers emerge. Acceptable as-is for now.

**P2-3: `rateLimiter.Wait()` returns `ctx.Err()` on success path**
- **File:** `internal/backfill/ratelimit.go:73,82`
- **What:** Both the "unlimited" early-return and the "tokens acquired" path return `ctx.Err()`. When ctx is not cancelled, this returns nil (correct). But the idiom is unusual — a reader might think "does this always error?"
- **Impact:** Readability only. Functionally correct.
- **Diff-scope:** New code.
- **Suggestion:** Consider `if err := ctx.Err(); err != nil { return err }; return nil` for clarity.

**P2-4: `writeMain` slice comment slightly misleading**
- **File:** `internal/backfill/runner.go:253`
- **What:** `next := remaining[:0:0]` comment says "新底层数组" (new underlying array), but the slice initially shares the underlying array with `remaining`. Since cap=0, the first `append` allocates fresh — so the *effective* behavior is correct, but the comment overstates.
- **Impact:** Nil — behavior is correct.
- **Diff-scope:** New code.

**P2-5: `build_os_image` in harness builds Docker image twice**
- **File:** `harness/run-backfill.sh:55-58`
- **What:** Two consecutive `docker build -q` calls. The first sends stdout to stderr (progress); the second's stdout is captured. Second build is cached (instant), so harmless.
- **Impact:** Negligible — harness script, not production.
- **Diff-scope:** New code.

## 3. Suggestions

1. **Add `fsyncDir` to checkpoint persist** (P2-1): One-line addition for consistency with the more careful DLQ path.
2. **Consider a shared `safeTableName`** (P2-2): If a third consumer beyond backfill/recon/harness emerges, extract to `internal/safename`.
3. **Clarify `rateLimiter.Wait` return** (P2-3): Minor readability improvement.

## 4. Additional Observations

- **Test coverage is thorough**: 343 lines of DLQ tests covering crash recovery (dirty suffix, corruption in synced prefix, no-sidecar, file-shorter-than-sidecar), idempotency, windowed counting, and replay-on-reopen. Runner tests cover happy path, idempotent, raw_excluded, DLQ routing, transient retry, batch-level retry, resume, multi-table, and ctx cancellation.
- **The `ReconcileChecked` migration in `cmd/reconcile`** is a genuine safety improvement: self-inconsistent inputs (DLQ > source_rows, raw_excluded > es_docs) are now rejected before arithmetic, preventing false OK on bad data.
- **The reconcile equation is clean**: `ES_docs == source_rows - DLQ`. `raw_excluded` correctly occupies an ES doc (content=null) and is NOT subtracted. This is well-documented and tested.
- **The harness e2e** (`run-backfill.sh` + `harness/backfill/main.go`) provides real-infra validation with a controlled 6-row suite, positive assertion of DLQ'd row absence from ES, and IK recall verification.

## 5. Data Flow Tracing

|---|---|---|
| `row.Signal`, `row.Setting` | `MySQLSource.ReadBatch()` → SQL `SELECT ... signal, setting ...` | ✅ Both checked in `isSignalEncrypted()` |
| `row.Payload` | Same SQL → `payload BLOB` column | ✅ Unmarshalled in `extractMessage()`; nil/empty → DLQ |
| `row.CreatedUnix` | `UNIX_TIMESTAMP(created_at)` in SQL | ✅ Stored in `dlqRecord.CreatedAt` for windowed counting |
| `row.MessageID` | SQL `message_id` column | ✅ Used as ES `_id` (idempotent upsert) and DLQ dedup key |
| DLQ `seen` map | Rebuilt from spill file on reopen via `recoverAndLoad()` | ✅ Truncated to synced offset; dedup preserved across restarts |
| `dlq.CountInWindow(from,to)` | Iterates `seen` values (created_at) | ✅ Aligns with recon `rangeFilter` gte/lte semantics |
| Checkpoint `LastID` | `Advance()` after `dlq.Sync()` succeeds | ✅ Monotonic guard; atomic persist; rollback on failure |
| Reconcile `sourceRows` | `MySQLSourceCounter.CountRows()` — `COUNT(*) WHERE UNIX_TIMESTAMP(created_at) BETWEEN ? AND ?` | ✅ Same window as DLQ and ES counts |
| Reconcile `esDocs` | `OSCounter.CountDocs()` — range query on `created_at` | ✅ Force refresh before count; shard completeness checked |
| `BulkItemResult.Status` | `esindex.Writer.Bulk()` per-item response | ✅ `Permanent()` = 4xx (not 429) → DLQ; transient → retry |

- **C1 — Dual-path parity:** N/A. This PR adds a new path (backfill) that is explicitly designed to be parallel-safe with the live Kafka increment via `_id=message_id` idempotent upsert. There is no add/remove or subscribe/unsubscribe symmetry to check. The two paths write the same ES doc shape via the same `internal/esindex.Writer`.
- **C2 — Control-flow ordering / nested reuse:** Clear. The durability ordering (DLQ Sync → Checkpoint Advance) is enforced at a single call site (`runTable`). The `extractMessage` function is called once per row with no nested/compound invocation. The `ReconcileChecked` wraps `Reconcile` (not the other way around) — no double-application risk.
- **C3 — Authorization boundary ≠ capability boundary:** N/A. This is an ops tool (`cmd/backfill`), not a user-facing endpoint. It reads message tables (read-only) and writes to ES (same index as the live indexer). Access is controlled by MySQL DSN and ES credentials, both requiring explicit configuration.

---

**[Octo-Q] verdict: APPROVE**

No P0/P1 findings. The PR is well-architected with strong attention to crash recovery (durability ordering, dirty-suffix truncation, idempotent dedup), reconcile gate integrity (self-consistency validation, windowed DLQ counting), and payload extraction alignment (pinned by shared test vectors). Five P2 findings are all maintainability/readability nits with no correctness impact. The test suite is comprehensive (unit + e2e harness) and covers the critical crash-recovery and idempotency paths.

Jerry-Xin (approved):
The PR is in scope for `octo-search-indexer`: it extends the existing message indexing pipeline with a backfill tool, reuses `internal/esindex`, and hardens the reconcile gate.

**💬 Non-blocking**

🟡 Warning: [internal/backfill/checkpoint.go](/Users/mlamp/.openclaw/tmp/pr-review-reqv_s9x/repo/internal/backfill/checkpoint.go:148) writes the checkpoint via temp file + `rename`, but does not `fsync` the parent directory afterward. This is not a correctness blocker because replay is idempotent and DLQ deduplication is durable, but after a host crash the checkpoint directory entry can be lost or revert, causing unnecessary replay despite the “durable checkpoint” expectation.

🔵 Suggestion: [harness/run-backfill.sh](/Users/mlamp/.openclaw/tmp/pr-review-reqv_s9x/repo/harness/run-backfill.sh:40) readiness loops do not fail if OpenSearch/MySQL never become ready; they fall through and print “stack up.” Make the loops assert readiness explicitly for clearer harness failures. Also, [harness/run-backfill.sh](/Users/mlamp/.openclaw/tmp/pr-review-reqv_s9x/repo/harness/run-backfill.sh:47) builds the IK OpenSearch image twice.

**✅ Highlights**

- Backfill correctly reuses `internal/esindex.Writer` rather than duplicating ES bulk logic.
- Source scanning uses keyset pagination and validates dynamic table names.
- DLQ spill accounting is thoughtfully designed: deduped replay, synced-offset recovery, windowed counts, and fail-closed behavior.
- `ReconcileChecked` improves gate safety by rejecting obviously inconsistent counts.
- Unit coverage is strong around DLQ recovery, checkpointing, extraction parity, retry handling, and reconcile arithmetic.

Verification: `go test ./...` passed.

yujiawei (commented):
# Code Review — PR #8 (octo-search-indexer)

**Reviewed at head SHA** `72ee8fb66d5002a43041062ad09410571346b7bd` against merge-base `c7b4b459` (`origin/main`). Build, `go vet`, `go test ./...`, and `go test -race ./internal/backfill/...` all pass locally.

This review used a three-perspective process (line-level adversarial pass, long-context whole-repo pass, and a structured multi-dimension fan-out with independent verification of every candidate finding). Findings below are deduped and severity-normalized; each was re-checked against the actual code before inclusion.

## 1. Summary verdict

**APPROVED.** This is a careful, well-tested addition. The backfill job correctly reuses `internal/esindex` (no re-implemented write path), uses keyset pagination (never `OFFSET`), and the durability story is genuinely solid: `DLQ.Sync()` (content fsync → atomic offset-sidecar advance, with parent-dir fsync) is ordered strictly *before* `checkpoint.Advance()`, so the high-water mark can never lead the durable DLQ state. The crash-recovery model (truncate the un-fsynced dirty suffix back to the synced offset, then strictly parse the clean `[0, offset)` prefix) is the right design and is well covered by tests. The reconcile equation `ES_docs == source_rows − DLQ` is sound, and the new `ReconcileChecked`/`Validate` self-consistency gate (reject `DLQ > source_rows`, `raw_excluded > es_docs`, negatives) is a good fail-closed hardening.

No correctness, data-loss, security, or build-blocking issue survived verification. **All findings below are P2 / advisory** — none should block merge. Several are worth a quick follow-up because this tool will eventually run against a 3.15M-row production dataset.

### Note on the two findings that initially looked like P1

Two candidates were raised as potential data-loss P1s and **both were downgraded after verification**, because the durability ordering neutralizes them:

- **`DLQSpill.Close()` swallows a post-fsync sidecar-write failure** — if `f.Sync()` succeeds but `writeSyncedOffset()` fails, the just-synced records get truncated as a "dirty suffix" on the next open. This is **not permanent loss**: `Sync()` runs *before* `checkpoint.Advance()` for every batch, so any record whose offset wasn't recorded belongs to a batch whose checkpoint was never advanced — resume re-reads those source rows from MySQL and re-spills them. Conservative-truncate + resume-rescan is the intended invariant. (P2: still worth surfacing the error to a non-zero exit so operators know a flush failed.)
- **`checkpoint.go persistLocked()` lacks the parent-dir fsync** that `dlq.go` has after its rename. Worst case is the checkpoint *reverting* after a host crash, causing idempotent re-processing (`_id=message_id` upsert dedups, DLQ `Write` is idempotent). No loss — but the asymmetry is real and easy to close.

## 2. Findings (all P2 / advisory)

**P2-1 — `Bulk` result-cardinality is assumed, not asserted, in `writeMain`.**
`internal/backfill/runner.go:193-224`. The per-item loop indexes `remaining[i]` over `results` with no `len(results) == len(remaining)` guard. The current `esindex.Writer` *does* guarantee an equal-length, order-aligned slice (`mapBulkResults`/`batchFailure` both build `make([]…, len(msgs))`, and a missing response item is marked transient — verified out-of-diff), so this is not a live bug. But the contract is load-bearing for "no silent skip"; a one-line cardinality check (treat a mismatch as a batch-level transient) makes the safety property local instead of dependent on a separate package.

**P2-2 — `checkpoint.go` rename is not directory-fsynced (asymmetry with `dlq.go`).**
`internal/backfill/checkpoint.go:148-152` renames the temp file but never `fsyncDir`s the parent, whereas `dlq.go:317-323` does after its sidecar rename. Only consequence is a possible checkpoint revert → idempotent reprocessing, so not blocking, but mirroring `fsyncDir(dir)` here restores the stated "checkpoint persisted" guarantee for one line of code.

**P2-3 — `Close()` / final-flush errors don't reach the process exit code.**
`cmd/backfill/main.go:124-128` logs `dlq.Close()` errors as `warn:` and the deferred close cannot change the exit status. Combined with P2-1's note, an `ENOSPC` on a final flush would log-but-exit-0. Given fail-closed is the design intent everywhere else, a flush/close failure should produce a non-zero exit.

**P2-4 — inline reconcile `-to` defaults to `now()`, which races the live pipeline.**
`cmd/backfill/main.go:160-163`. Because backfill runs alongside the live Kafka indexer, an open-ended `-to = now()` lets `source_rows` include in-flight rows not yet in ES → false `MISMATCH`. The runbook example does use explicit `-from/-to` epoch bounds, so this is a footgun on the default rather than a bug — consider requiring an explicit `-to` (or snapshotting a per-table `max(id)` cutoff at job start) when `-reconcile` is set.

**P2-5 — open-ended scan has no per-table upper bound.**
`internal/backfill/source.go:39-42` (`WHERE id>? ORDER BY id ASC LIMIT ?`). On hot shards the scan can chase rows inserted during the run, making the "historical" window operationally fuzzy. If the live path is meant to own the moving tail, capture `max(id)` per table at start. Same root cause as P2-4.

**P2-6 — DLQ dedup key is `message_id`, but reconcile counts source *rows*.**
`internal/backfill/dlq.go:181-186`. If two distinct source rows ever shared a non-empty `message_id` and both DLQ'd, only one would be counted. This fails *closed* (ES also dedups by `_id=message_id`, so duplicates break the equation → MISMATCH, not a false OK), and `message_id` is effectively the unique IM message key. Worth a one-line doc note stating the global-uniqueness assumption the accounting relies on.

**P2-7 — non-text / unknown-shape payloads become `raw_excluded` docs the gate can't catch.**
`internal/backfill/extract.go:107-123`. A payload like `{"content":"hello"}` (no/!=Text `type`) becomes an ES doc with `content=null`, and the reconcile equation still balances. This is intended (mirrors the producer's conservative split), and the test pins the vectors against the producer — but it does mean the gate validates *count*, not *content fidelity*. No action needed if producer parity holds; flagging because it's the one class of "loss" the gate is blind to by design.

**P2-8 — no `db.SetConnMaxLifetime`.**
`cmd/backfill/main.go:88-99`. A long/paused run can outlive MySQL `wait_timeout` and hit `invalid connection` on the next keyset query. Setting a conn max-lifetime (and/or relying on the driver's retry) avoids a mid-run abort on a multi-hour job.

### Minor / nits (non-blocking)
- `expJitterBackoff` can return 0 (full jitter includes 0) → an occasional immediate retry. Harmless; matches the consumer's semantics.
- `writeMain` per-item transient retry is unbounded (by design — bounded by ctx timeout/SIGTERM). Fine for a backfill, but a sustained ES 5xx will spin until the 6h timeout; an attempt cap or clearer "stuck" log line would aid operability.
- The single 6h context timeout covers both the backfill *and* the inline reconcile; a very slow reconcile shares the same deadline. Minor.
- `rateLimiter` has no mutex, but `Wait` is only ever called from the single runner goroutine (`-race` clean), so there is no live race. Add a mutex only if it ever becomes shared.

## 3. Coverage notes (what each angle could and couldn't verify)

- **Verified in-repo (beyond the diff):** `esindex.Writer.Bulk` length/ordering guarantee and permanent-vs-transient classification; reconcile window inclusivity is consistent across all three counters (MySQL `BETWEEN`, ES `gte/lte`, DLQ `>=/<=`, all on `created_at` epoch seconds); single-goroutine runner; full suite + `-race` green.
- **Not verifiable from code here (assumptions to confirm operationally):** the real `message` table schema constraints (global `message_id` uniqueness, nullable columns, session timezone for `UNIX_TIMESTAMP`); that the copied `contentTypeText`/Signal-bit constants still match the live producer (the test pins this, but a producer-side change is an external dependency); and actual behavior under a real 3.15M-row run (memory of the in-RAM `seen` map is fine for the expected tiny DLQ volume, but would grow unbounded under a systemic payload bug — see P2-7 / robustness).

## 4. Recommendation

Approve. The P2 items are real but none gate merge; P2-1, P2-2, and P2-3 are the cheapest high-value follow-ups (they harden the fail-closed contract this design already commits to), and P2-4/P2-5 are worth deciding before the first production run so the reconcile gate doesn't false-alarm against the live pipeline.

Jerry-Xin (approved):
The PR is relevant to this repository and the implementation is generally sound: it adds an offline message backfill path that reuses the existing OpenSearch writer and strengthens reconcile accounting.

💬 Non-blocking

🟡 Warning — `cmd/backfill/main.go:147` and `cmd/backfill/main.go:159`: `-reconcile` is documented as requiring `-from/-to`, but `from` defaults to `0`, so an operator can accidentally reconcile from Unix epoch to now. Consider rejecting `-reconcile` with `from <= 0` unless an explicit “all history” mode is intended.

🔵 Suggestion — `internal/backfill/runner.go:192-224`: `writeMain` assumes `writer.Bulk` returns a result slice exactly aligned with the input. The current `internal/esindex` writer does, but a defensive length check would better preserve the backfill’s fail-closed posture if another implementation or future bug returns too few or too many results.

🔵 Suggestion — `harness/run-backfill.sh:47-50`: `build_os_image` runs the same `docker build` twice. Capture the image ID once and print/log it if needed; this will make the harness faster and less noisy.

🔵 Suggestion — `harness/run-backfill.sh:64-78`: the standalone `backfill` and `verify` subcommands rely on `BASE_TS` being set by `seed` in the same shell. Running them independently under `set -u` fails unless the caller manually exports `BASE_TS`. Either persist the timestamp from `seed` or document that standalone subcommands require `BASE_TS`.

✅ Highlights

- Good reuse of `internal/esindex.Writer`; no duplicate OpenSearch write path.
- Keyset scanning, table-name allowlisting, checkpoint isolation, and DLQ fsync-before-checkpoint ordering are appropriate for this job.
- Reconcile hardening with `ReconcileChecked` is a useful guard against impossible accounting inputs.
- Test coverage is strong around DLQ replay, idempotency, windowed counting, checkpoint persistence, and runner behavior.

Verification run: `go test ./...` and `go test -race ./internal/backfill ./internal/recon` both passed.

mochashanyao (approved):
[Octo-Q · automated review]

---

# octo-search-indexer PR#8 Review Report
**PR:** Mininglamp-OSS/octo-search-indexer #8
**Head SHA:** `e7ca0814390e1f358c855cc3cf8273efa70e14e6`
**Title:** feat(backfill): historical message backfill job + DLQ-accounting reconcile gate

---

## 1. Verification Summary

| Area | Status | Evidence |
|------|--------|----------|
| Keyset pagination (no OFFSET) | ✅ | `source.go:48` — `WHERE id>? ORDER BY id ASC LIMIT ?` with parameterized args |
| SQL injection prevention | ✅ | `source.go:47` + `source.go:72-85` — `safeTableName` allowlist `[A-Za-z0-9_]` + backtick quoting |
| ES write reuse (internal/esindex) | ✅ | `runner.go:82` — `r.writer.EnsureIndex(ctx)`; `runner.go:170` — `r.writer.Bulk(ctx, msgs)` |
| Idempotent upsert (`_id=message_id`) | ✅ | `esindex/writer.go:119` — `action.Index.ID = msgs[i].MessageID`; index action (not create) = upsert |
| DLQ durability ordering | ✅ | `runner.go:119-126` — `dlq.Sync()` before `cp.Advance()`; fail-closed on either error |
| DLQ crash recovery | ✅ | `dlq.go:73-98` — sidecar offset truncates un-fsynced dirty suffix; clean prefix parse is strict |
| DLQ idempotent dedup | ✅ | `dlq.go:186-192` — `seen` map keyed by `dedupKey(rec)`; duplicate Write is no-op |
| Checkpoint atomic write | ✅ | `checkpoint.go:107-157` — temp file + fsync + rename + fsyncDir |
| Checkpoint memory rollback | ✅ | `checkpoint.go:91-93` — `s.cp.LastID[table] = prev` on persist failure |
| Payload extraction alignment | ✅ | `extract.go:36-42` — constants `contentTypeText=1`, `signalSettingMask=1<<5`; pinned by `extract_test.go:39-45` |
| Reconcile gate (ReconcileChecked) | ✅ | `recon.go:56-78` — Validate() rejects negative, DLQ>SourceRows, RawExcluded>ESDocs |
| Inline reconcile force-refresh | ✅ | `cmd/backfill/main.go:175-178` — `Indices.Refresh` before `_count` |
| Windowed DLQ count | ✅ | `dlq.go:213-222` — `CountInWindow` iterates `seen` map with `gte/lte` bounds |
| Fail-closed on empty spill dir | ✅ | `dlq.go:68-70` — `OpenDLQSpill("")` returns error |
| Context cancellation propagation | ✅ | `runner.go:99-101` — `ctx.Err()` checked at loop top; `ratelimit.go:68` — returns `ctx.Err()` |
| Transient retry (per-item + batch) | ✅ | `runner.go:155-199` — batch-level retries whole batch; per-item 429/5xx stays in `remaining` |
| Test coverage | ✅ | 20+ unit tests covering happy path, DLQ, resume, crash recovery, idempotency, multi-table |

## 2. Findings

### P2-1: Stats.DLQ overcounts on resume after crash (cosmetic)

In `runner.go:143-148`, `processBatch` increments `s.DLQ++` unconditionally after `dlq.Write()`. But `dlq.Write` is idempotent — if the same message_id was already spilled (from a prior run that crashed after DLQ write but before checkpoint advance), the write is a no-op while `s.DLQ` still increments.

**Impact:** `Stats.DLQ` in the log line may be higher than `dlq.Count()` after a resume. This is **cosmetic only** — the reconcile gate uses `dlq.CountInWindow()` (authoritative), not Stats.

**Path:** `runner.go:143-148` (processBatch) → `dlq.go:186-192` (Write idempotent no-op)

**Suggestion:** Guard the increment: `if !dlq.Seen(key) { s.DLQ++ }` or have `Write` return a bool indicating whether it was a new record.

### P2-2: `next := remaining[:0:0]` comment is misleading about mechanism

**Diff-scope:** new

`runner.go:178`: `next := remaining[:0:0]` — the comment says "新底层数组，避免别名污染" but `[:0:0]` does not allocate a new array; it creates a zero-len/zero-cap slice sharing the old backing array. The **next `append`** allocates a new array because cap=0. The behavior is correct (no aliasing after append), but the comment misattributes the mechanism.

**Path:** `runner.go:178`

**Suggestion:** Rewrite comment to: `// cap=0 forces append to allocate a new backing array on first use`.

### P2-3: `Validate()` has a redundant check

**Diff-scope:** new (amplified by this PR)

`recon.go:72-74` checks `SourceRows - DLQ < 0`, which is mathematically equivalent to `DLQ > SourceRows` (already checked at line 66-68) given all values are non-negative (checked at line 63-65). The fourth check is unreachable if the second passes.

**Path:** `recon.go:63-75`

**Suggestion:** Remove the redundant check or add a comment noting it's defense-in-depth.

### P2-4: `safeTableName` duplicated across 3 locations

**Diff-scope:** new

Defined in `internal/backfill/source.go:72-85`, `harness/backfill/main.go:260-273`, and exists in `internal/recon` (pre-existing). Three copies of the same allowlist logic.

**Suggestion:** Consider extracting to a shared internal utility. Low priority — the function is trivial and the harness is a separate binary.

### P2-5: `build_os_image` in harness builds the same Docker image twice

**Diff-scope:** new

`harness/run-backfill.sh:53-56` runs `docker build -q` twice with identical arguments. The first sends output to stderr, the second to stdout (for command substitution). The second build is a cache hit but still incurs overhead.

**Suggestion:** Capture the image ID once: `local img; img=$(docker build -q ...); echo "$img" >&2; echo "$img"`.

## 3. Suggestions

- **P2-1 fix** (Stats overcount): Low effort — have `DLQSpill.Write` return `(isNew bool, err error)` so callers can conditionally increment.
- **Consider adding a max retry cap** for transient per-item retries in `writeMain`. Currently bounded only by context timeout (default 6h). A per-item retry counter with a "give up and DLQ" threshold would prevent a single flaky item from consuming the entire timeout budget. Not blocking — the timeout does bound it.

## 4. Additional Observations

- **No production Dockerfile change** — `cmd/backfill` is an ops-only binary, not baked into the service image. ✅ Good isolation discipline.
- **Harness e2e is well-designed** — 6-row seeded suite with obvious gate arithmetic (6 source − 1 DLQ = 5 ES docs), positive assertion that DLQ'd row is absent from ES, and IK recall verification.
- **Commit history shows iterative hardening** — 7 commits progressing from initial implementation through crash recovery fixes (sidecar offset, dirty suffix truncation, fsyncDir after rename). Each fix has a corresponding test.

## 5. Data-Flow Trace

| Consumed datum | Upstream source | Traces correctly? |
|---|---|---|
| `srcMessageRow.ID` | SQL `SELECT id` → `rows.Scan(&r.ID)` | ✅ keyset cursor + checkpoint |
| `srcMessageRow.MessageID` | SQL `SELECT message_id` → `rows.Scan(&r.MessageID)` | ✅ used as ES `_id` + DLQ dedup key |
| `srcMessageRow.Payload` | SQL `SELECT payload` → `rows.Scan(&r.Payload)` | ✅ fed to `json.Unmarshal` in extract |
| `srcMessageRow.Setting` | SQL `SELECT setting` → `rows.Scan(&r.Setting)` | ✅ bit-5 checked for Signal |
| `srcMessageRow.Signal` | SQL `SELECT signal` → `rows.Scan(&r.Signal)` | ✅ non-zero = Signal encrypted |
| `srcMessageRow.CreatedUnix` | SQL `UNIX_TIMESTAMP(created_at)` → `rows.Scan(&r.CreatedUnix)` | ✅ stored in DLQ record for windowed count |
| `extractMessage` outcome | `isSignalEncrypted(row)` → `json.Unmarshal(row.Payload)` → `payloadType(m)` → `m["content"].(string)` | ✅ 3-way split aligned with producer |
| `BulkItemResult.OK/Status` | `mapBulkResults` aligns by index with input `msgs` | ✅ permanent → DLQ, transient → retry |
| `dlq.CountInWindow` | `seen` map (message_id → created_at), rebuilt from spill on reopen | ✅ authoritative for reconcile |
| `cp.Get(table)` | `Checkpoint.LastID[table]`, loaded from JSON or 0 | ✅ monotonic advance only |
| `ReconcileChecked` inputs | `srcCounter.CountRows` / `esCounter.CountDocs` / `esCounter.CountRawExcluded` / `dlq.CountInWindow` | ✅ all windowed, force-refreshed |

## 6. R5 Blindspot Checklist

- **C1 — Dual-path parity:** N/A. This PR adds a new standalone tool (`cmd/backfill`); there is no symmetric add/remove or subscribe/unsubscribe pair. The backfill path and live-ingest path are explicitly designed as independent, idempotent writers sharing `_id=message_id`. No parity concern.

- **C2 — Control-flow ordering / nested reuse:** Clear. The durability ordering (DLQ Sync → checkpoint Advance) is enforced at a single call site (`runTable`). `ReconcileChecked` wraps `Reconcile` with validation — no double-application risk. `safeTableName` is a pure function with no side effects. No nested-reuse concern.

- **C3 — Authorization boundary ≠ capability boundary:** N/A. `cmd/backfill` is an ops CLI tool requiring explicit MySQL DSN + ES credentials via flags/env. It is not an API endpoint. No authorization boundary to audit.

---

**[Octo-Q] verdict: APPROVE**

No P0 or P1 findings. All 5 findings are P2 (cosmetic/maintainability). The code demonstrates strong durability engineering (fsync ordering, crash recovery via sidecar offset, idempotent dedup), correct data-flow alignment with the live producer, and comprehensive test coverage including crash-recovery scenarios. The reconcile gate hardening (`ReconcileChecked` + `Validate`) is a sound defensive improvement.

yujiawei (commented):
# Code Review — PR #8 (octo-search-indexer)

**Scope:** Phase-6 historical message backfill job (`cmd/backfill` + `internal/backfill/*`) and the hardened reconcile gate (`internal/recon`, `cmd/reconcile`). Reviewed at head `e7ca0814390e1f358c855cc3cf8273efa70e14e6` against merge-base `c7b4b459`. `go build`, `go vet`, `go test ./internal/...` (incl. `-race` locally) all green; 35 new tests cover resume, dedup, dirty-suffix recovery, no-double-count, and reconcile self-consistency.

This review combined a manual read with multiple independent adversarial passes (4 dimension-specialized legs + a cross-model leg), each finding re-verified against the actual code. Both auxiliary legs ran successfully.

## Verdict

**APPROVED.** No P0/P1 (correctness / data-loss / security / build-break) issues survive verification. The design is sound: the `_id=message_id` idempotency premise makes the bypass-Kafka path safe alongside the live increment; the DLQ-spill durability machinery (synced-offset sidecar, fsync-before-checkpoint ordering, dedup-by-message_id, dirty-suffix truncation on resume) is correct and well-tested; the reconcile gate is fail-closed in the right direction (it reports MISMATCH conservatively, never a false OK on a real shortfall). All findings below are P2 hardening/robustness items — none blocks merge.

## 1. Validation summary

- ✅ **Idempotency / dual-path safety** — `_id=message_id` upsert; backfill + live overlap is safe. (`internal/backfill/extract.go`, reuse of `internal/esindex`)
- ✅ **Keyset pagination** — `WHERE id>? ORDER BY id ASC LIMIT ?`, no OFFSET, monotonic checkpoint advance only after a batch is terminal. (`source.go:39-42`, `runner.go:118-128`)
- ✅ **DLQ durability ordering** — `dlq.Sync()` (fsync content + atomic sidecar) strictly before `cp.Advance()`. (`runner.go:118-127`, `dlq.go:258-283`) Crash-resume re-reads un-advanced ids; ES upsert + message_id dedup make replay harmless. Verified by `TestRunner_ResumeAfterDLQNoDoubleCount`.
- ✅ **Reconcile gate** — `Validate()` rejects self-inconsistent inputs (DLQ>source, raw_excluded>es_docs, negatives); windowed DLQ count (`CountInWindow`) matches the recon range filter; `_count` rejects partial-shard results. Fail-closed: false OK is not reachable via these inputs.
- ✅ **SQL safety** — every dynamic table name (read path, recon, harness DROP/CREATE) is allowlist-validated `[A-Za-z0-9_]` and backtick-wrapped; time/paging via placeholders. (`source.go:71-86`, `collectors.go:59-69`, `harness/backfill/main.go:264-279`)
- ✅ **Producer alignment** — extract three-way split pinned to producer vectors + the two octo-lib constants by `extract_test.go` (any drift turns the test red). See blind-spot note below.

## 2. Findings (all P2 — hardening, non-blocking)

**Consensus across legs (highest confidence):**

- **P2 — reconcile window vs live traffic (default `-to=0`→now).** `cmd/backfill/main.go:160-163` sets the upper bound to `time.Now().Unix()` when `-to` is unset. Against an actively-written shard, the newest source rows are still in-flight through the live Kafka path and not yet searchable, so `SourceRows > ESDocs` yields a spurious `MISMATCH`. This is fail-closed (a false *mismatch*, never a false OK) and the runbook already prescribes explicit `-from/-to`, so it does not block — but the default is a footgun. Suggest either requiring an explicit `-to`, or applying a live-latency buffer (e.g. `now - 300s`) and documenting it. *(One leg rated this P1; downgraded to P2 here because the direction is conservative and the operator workflow uses an explicit window — but it is the top item to address.)*
- **P2 — source scan is not NULL-tolerant.** `source.go:56-61` scans `signal`/`setting`/`channel_type`/`timestamp`/`created_unix` into plain non-nullable Go types; a NULL in any of these aborts the whole batch read, where the dbr-based live producer tolerates it. For a one-shot historical load over 3.15M legacy rows this is a realistic stop. Consider `sql.NullInt64`/`COALESCE` for the columns that can legally be NULL in old data.

**Unique (single-leg, verified real):**

- **P2 — `writeMain` trusts `Bulk` result alignment with no guard.** `runner.go:200-201` does `for i, res := range results { item := remaining[i] }` assuming `len(results)==len(remaining)`. Today `esindex.Bulk` always returns a slice of `len(msgs)`, so it holds — but the *live consumer* defensively guards the same call (`if j < len(bulkRes)` in `consumer.go`). Mirror that guard so a future writer-contract change can't silently drop rows past an advancing checkpoint (or panic).
- **P2 — `writeMain` retries any non-OK / non-permanent status forever.** `runner.go:220-222` routes everything that isn't 2xx or `Permanent()` (4xx≠429) to the `default` in-place-retry bucket, bounded only by the 6h global timeout. ES `_bulk` index actions don't return 3xx, so this isn't reachable today and matches the live consumer's pattern — but a status-0-without-batch-error or any odd status would hang the job. A max-attempts escalation (alert / fail) would harden it.
- **P2 — `DLQSpill.Write` lacks the `s.f == nil` guard** that `Sync`/`Close` have (`dlq.go:221`). Single-goroutine usage means no live bug, but a `Write` after the deferred `Close` would nil-panic instead of failing cleanly. Add `if s.f == nil { return fmt.Errorf("dlq spill closed") }`.
- **P2 — `recoverAndLoad` can index `data[-1]` on a 0-byte TOCTOU.** `dlq.go:164` evaluates `data[len(data)-1]` after `os.ReadFile`; if an external process truncates the spill to 0 bytes between the `os.Stat` and the read (with `syncedLen>0`), `len(data)==0` panics on restart. Add a `len(data)==0` check that returns the same "durability state lost" error used elsewhere.
- **P2 — recovery truncate is not fsync'd.** `dlq.go:147-153` truncates the dirty suffix but doesn't fsync the file/dir, and logs "recovered" before it's durable. It self-heals via the sidecar (next reopen re-truncates), so it's not data loss — but an `fsync`-after-truncate or a one-line comment stating the truncate is intentionally non-durable would close the gap.
- **P2 — orphaned temp files never swept.** `writeSyncedOffset` (`dlq.go:288`) and `persistLocked` (`checkpoint.go:119`) create `.tmp` files cleaned only on in-function error paths; a SIGKILL between `CreateTemp` and `Rename` orphans one per crash, and reopen never globs them. Add a startup sweep of stale `.backfill-cp-*.tmp` / `.backfill-dlq-synced-*.tmp` (safe — live files are atomic-renamed).
- **P2 — `Stats.DLQ` double-counts on resume (cosmetic).** `runner.go:159-160,216-217` increment the counter even when `dlq.Write` is a dedup no-op, so the logged `dlq=` can exceed the true deduped count after a resume. The reconcile gate uses `CountInWindow` (the authoritative deduped count), not `Stats.DLQ`, so correctness is unaffected — fix only to keep the log honest.
- **P2 — inline reconcile ignores `_refresh` shard failures.** `cmd/backfill/main.go:176-178` checks only `rerr != nil`, not `resp.Shards.Failed`. The subsequent `_count` does check shards, so a partial-refresh impact is limited, but checking the refresh response too would be consistent.
- **P2 — `fsyncDir` is POSIX-only.** `dlq.go:189` `d.Sync()` on a directory handle errors on Windows; local `cmd/backfill`/harness runs on a Windows host would hard-fail at init. No-op it for `GOOS=windows` if local Windows dev is in scope (likely not for a Linux ops job — noting for completeness).

## 3. Recommendations (priority order)

1. Harden the reconcile default window (require `-to`, or apply+document a live-latency buffer) — the one item with real operational bite.
2. Make the source scan NULL-tolerant for legacy columns.
3. Add the `len(results)` guard in `writeMain` and the `s.f==nil` guard in `Write` (cheap defensive parity with the consumer).
4. The remaining items (temp-file sweep, fsync-after-truncate comment, 0-byte guard, Stats log, refresh-shard check) are low-effort polish; batch at convenience.

## 4. Blind spots / coverage

- E2E verified only at toy scale (6 rows in `harness/`); the 3.15M-row / ~2.14GB production load, `refresh_interval=-1` behavior during a long backfill, and rate-limiter under real ES backpressure are not exercised here — expected, since the runbook makes a real run a separate signed-off low-peak ops action.
- Producer alignment is pinned by mirrored test vectors + the two octo-lib constants, but the actual `octo-server/modules/searchetl/payload.go` source is not in this repo, so the guarantee is "the vectors stay in sync," not a live cross-repo diff. If the producer changes its split, only the duplicated test catches it.
- The cross-model leg's diff was truncated (>2000 lines), so its coverage of `runner.Run`, `NewMySQLSource`, and the extract helpers came from the dimension legs + manual read, not that leg.

PR #10, opened 2026-06-16
feat(esindex): converge OS doc contract to reader shape + recon sampling

### LINKED ISSUE #11: Converge OS index doc to reader contract + fail-closed visibility handling
## Summary

Converge the OpenSearch document the `es-indexer` writes with the **octo-server
reader contract** so the read path can consume what this service indexes, and
harden the safety/visibility handling that gates message search results.

## Scope

- Index doc shape aligned to the reader (camelCase + nested `payload.*`,
  full-precision `messageId` long, reader-gated `spaceId`/`visibles`/`messageSeq`).
- Backfill enrichment self-sources `spaceId`/`visibles` from the raw message
  payload and `messageSeq` from the column.
- **Visibility ACL is fail-CLOSED**: a row whose visibility cannot be trusted to
  parse is routed to the DLQ; a legitimate `visibles` list is never silently
  dropped (which would fail-OPEN the admin-only group-system-message gate).
- Live-ingestion safety gate refuses to start while the Kafka contract lacks the
  reader safety fields, plus a CI guard that turns red if a future contract bump
  enables ingestion without wiring those fields through.
- Field-level reconciliation sample gate (independent parser) so a dropped ACL
  surfaces as a mismatch rather than going undetected.
- Forward-migration script idempotency + default-safe index bootstrap (read
  alias is attached only after reconciliation, never on bare index create).

## Acceptance

- `go build`, `go vet`, `go test -race`, `golangci-lint run` all green.
- Reader can consume the indexed doc; visibility ACL never fails open from the
  backfill path.

yujiawei:
## Summary

Converge the OpenSearch document the `es-indexer` writes with the **octo-server
reader contract** (`modules/messages_search/source.go::Doc`) so the read path can
actually consume what this service indexes. The reader is unchanged; only the
write side moves.

Before this change the indexer wrote a flat snake_case doc
(`message_id`/`content`/…) while the reader reads camelCase nested
(`messageId`/`payload.text.content`/…) plus three fields the reader gates on
(`spaceId`, `visibles`, `messageSeq`) — so the read path could not consume the
index at all.

## Changes

- **mapping v1.9** (`internal/esindex/mapping/octo-message.json`): camelCase +
  nested `payload.*` + `spaceId`(keyword) + `visibles`(keyword[]) +
  `messageSeq`(long) + `messageId`(long, full-precision snowflake) + IK analyzers
  + the `wukongim-messages-read` alias.
- **`esindex.Doc` + `DocFromMessage`**: contract → reader doc. A non-numeric
  `message_id` is a **permanent (DLQ) error**, never silently coerced to `0`
  (the reader drops `messageId==0`).
- **backfill enrichment** (`docFromRow`): self-sources `spaceId`/`visibles` from
  the raw message payload and `messageSeq` from the column. The live consumer
  path cannot (the Kafka contract carries none of them).
- **live-ingestion safety gate**: the consumer refuses to start writing while the
  Kafka contract lacks the safety fields, so it never emits docs that would
  fail-OPEN the reader's group-message visibility gate. Existing docs are
  enriched by backfill meanwhile.
- **writer**: `Bulk` converts per-item; new `BulkDocs` for the backfill path.
- **reconciliation** (`cmd/reconcile`, `make run-recon`): adds field-level sample
  comparison (incl. `_source.messageId` precision) on top of count drift, a
  structured report with pinned failure thresholds and an exit-2 gate, and an
  optional push payload for the read-only drift gauges.
- **forward migration** (`scripts/forward-migrate.sh`,
  `docs/forward-migration-v1.9.md`): write new index → reindex/backfill → atomic
  alias switch, with rollback kept.
- harness + tests updated to the converged contract (numeric ids, nested fields).

## Verification

- [x] `go build ./...`, `go vet ./...` clean
- [x] `go test ./...` PASS (incl. `-race`)
- [x] independent model review PASS (2 findings raised + fixed: live-ingestion
  fail-open gate, recon `_source.messageId` drift)

No production rollout in this PR: real ingestion stays off until a later,
separately signed-off step. Filling the live path's safety fields requires an
additive contract bump + producer change shipped together at that step.

---

## R2 rework (round 2, cap=3)

Fixes the issues raised on the second review pass:

- **🔴 `extractVisibility` fail-OPEN (backfill ACL silently cleared)** — the strict
  struct meant a non-string `space_id` (number/object/upstream drift) failed the
  whole `Unmarshal` and returned empty `visibles`, silently dropping a legitimate
  access-control list. The reader treats empty `visibles` as **fail-OPEN** (the
  admin-only group system-message whitelist gate is removed). Now `space_id` is
  type-tolerant (string-only, else empty → reader p2p fail-closed) and visibility
  parsing is **fail-CLOSED**: a structurally broken row (non-object payload /
  non-array `visibles`) is routed to the DLQ, never written with empty `visibles`.
  The recon sample gate's parser is re-implemented **independently** of the write
  path and made type-tolerant, so it can actually catch a dropped ACL instead of
  being blind to the same bug. Tests cover non-string `space_id` keeping
  `visibles`, fail-closed on broken visibility, and the gate not being blind.
- **🟡 v2-gate latent fail-open CI test** — a reflection-based guard
  (`internal/esindex/v2gate_test.go`) that turns red if a future contract bump
  (`SchemaVersion>=2`) enables live ingestion without wiring
  `spaceId`/`visibles`/`messageSeq` through `DocFromMessage`.
- **🟡 `forward-migrate.sh` alias idempotency** — step3 atomically removes the
  read alias from **any** index (`index:"*"`, `must_exist:false`) before adding it
  to the new index, so it is idempotent and the alias is always single-pointing.
  The read alias was removed from the embedded mapping so a bare `PUT` /
  `EnsureIndex` is default-safe (alias attached only after reconciliation).
- **🟡 reconciliation** — README documents the mandatory `make run-recon` gate.

Closes #11

### DIFF
--- a/.github/workflows/check-sprint.yml
+++ b/.github/workflows/check-sprint.yml
@@ -33,7 +33,14 @@ name: Check Sprint
 
 on:
   pull_request_target:
-    types: [opened, synchronize, reopened]
+    # Include "edited" so the check re-runs when the PR description is updated
+    # (e.g. a developer adds "Closes #<issue>" after opening) — the reusable
+    # workflow resolves Sprint from the linked issue, so a body edit must
+    # re-trigger it. "ready_for_review"/"converted_to_draft" keep the draft-skip
+    # safe (without them a green "skipped" result cached on the draft SHA would
+    # persist after promotion and bypass the Sprint gate). Matches the usage
+    # documented in reusable-check-sprint.yml.
+    types: [opened, synchronize, reopened, edited, ready_for_review, converted_to_draft]
     branches: [main]
 
 permissions: {}
--- a/Makefile
+++ b/Makefile
@@ -28,16 +28,19 @@ test: ## 跑全部单测
 ##   RECON_ES_INDEX    目标索引/别名（默认 octo-message；切换后对 wukongim-messages-read）
 ##   RECON_TABLES      message 分表（默认 message,message1,message2,message3,message4）
 ## 可选：RECON_SAMPLE（抽样数，默认 200）、RECON_FROM/RECON_TO（窗，纪元秒）、
-##       RECON_DLQ（窗内已知 DLQ 数）、RECON_PUSH_URL/RECON_PUSH_TOKEN（回填 octo-server gauge）。
+##       RECON_DLQ（窗内已知 DLQ 数）、RECON_DLQ_SPILL_DIR（backfill 落下的 DLQ spill 目录；
+##       其 message_id 从字段级抽样门排除，避免合法 DLQ 行被误判 sample_missing）、
+##       RECON_PUSH_URL/RECON_PUSH_TOKEN（回填 octo-server gauge）。
 RECON_FROM ?= 0
 RECON_TO   ?= 0
 RECON_DLQ  ?= 0
+RECON_DLQ_SPILL_DIR ?=
 
 run-recon: ## 跑对账（人类可读摘要 + 退出码 gate）
-	$(GO) run ./cmd/reconcile -from $(RECON_FROM) -to $(RECON_TO) -dlq $(RECON_DLQ)
+	$(GO) run ./cmd/reconcile -from $(RECON_FROM) -to $(RECON_TO) -dlq $(RECON_DLQ) -dlq-spill-dir '$(RECON_DLQ_SPILL_DIR)'
 
 run-recon-json: ## 跑对账（结构化 JSON 报告，供机检 / 入库）
-	$(GO) run ./cmd/reconcile -from $(RECON_FROM) -to $(RECON_TO) -dlq $(RECON_DLQ) -json
+	$(GO) run ./cmd/reconcile -from $(RECON_FROM) -to $(RECON_TO) -dlq $(RECON_DLQ) -dlq-spill-dir '$(RECON_DLQ_SPILL_DIR)' -json
 
 ## ── 前向迁移（步骤 6 三步）─────────────────────────────────────────────────────
 ## 写新契约索引(v1.9) → 存量 reindex → alias 原子切换。详见 docs/forward-migration-v1.9.md。
--- a/Makefile
+++ b/Makefile
@@ -1,0 +1,45 @@
+# octo-search-indexer — build / test / recon entrypoints (YUJ-4534 / YUJ-4682)
+.DEFAULT_GOAL := help
+
+GO ?= go
+
+.PHONY: help build test vet run-recon run-recon-json migrate-forward
+
+help: ## 列出可用目标
+	@grep -E '^[a-zA-Z_-]+:.*?## .*$$' $(MAKEFILE_LIST) | sort | awk 'BEGIN{FS=":.*?## "}{printf "  \033[36m%-18s\033[0m %s\n", $$1, $$2}'
+
+build: ## 编译全部二进制
+	$(GO) build ./...
+
+vet: ## go vet
+	$(GO) vet ./...
+
+test: ## 跑全部单测
+	$(GO) test ./... | tail -40
+
+## ── 对账作业（步骤 5）─────────────────────────────────────────────────────────
+## ES doc-count vs MySQL 行数 + 抽样字段比对，检出多 doc/少 doc/字段 drift。
+## 对平退出 0；不对平退出 2（CI / cron gate 失败信号）。
+## 指标/阈值钉死在 internal/recon（doc_drift!=0 / sample_mismatch!=0 / sample_missing!=0 即不健康）。
+##
+## 必需配置（env 或 flag）：
+##   RECON_MYSQL_DSN   MySQL DSN，如 'user:pass@tcp(host:3306)/im_prod'
+##   RECON_ES          OpenSearch 地址（默认 http://localhost:9200）
+##   RECON_ES_INDEX    目标索引/别名（默认 octo-message；切换后对 wukongim-messages-read）
+##   RECON_TABLES      message 分表（默认 message,message1,message2,message3,message4）
+## 可选：RECON_SAMPLE（抽样数，默认 200）、RECON_FROM/RECON_TO（窗，纪元秒）、
+##       RECON_DLQ（窗内已知 DLQ 数）、RECON_PUSH_URL/RECON_PUSH_TOKEN（回填 octo-server gauge）。
+RECON_FROM ?= 0
+RECON_TO   ?= 0
+RECON_DLQ  ?= 0
+
+run-recon: ## 跑对账（人类可读摘要 + 退出码 gate）
+	$(GO) run ./cmd/reconcile -from $(RECON_FROM) -to $(RECON_TO) -dlq $(RECON_DLQ)
+
+run-recon-json: ## 跑对账（结构化 JSON 报告，供机检 / 入库）
+	$(GO) run ./cmd/reconcile -from $(RECON_FROM) -to $(RECON_TO) -dlq $(RECON_DLQ) -json
+
+## ── 前向迁移（步骤 6 三步）─────────────────────────────────────────────────────
+## 写新契约索引(v1.9) → 存量 reindex → alias 原子切换。详见 docs/forward-migration-v1.9.md。
+migrate-forward: ## 执行前向迁移（写新索引 + reindex + alias 切，禁半新半旧同 alias）
+	./scripts/forward-migrate.sh
--- a/README.md
+++ b/README.md
@@ -103,14 +103,22 @@ the field-level reconciliation gate after a backfill:**
 # inline after the backfill run (uses the job's own exact DLQ count as authority):
 go run ./cmd/backfill ... -reconcile -from <epoch> -to <epoch>   # count gate + field-level sample gate
 
-# or standalone, any time:
-make run-recon RECON_FROM=<epoch> RECON_TO=<epoch> RECON_DLQ=<dlq-count>
+# or standalone, any time (pass the backfill DLQ spill dir so legit DLQ rows are
+# excluded from the sample gate exactly like the inline path — otherwise a sampled
+# DLQ row false-fails as sample_missing and the gate exits non-zero):
+make run-recon RECON_FROM=<epoch> RECON_TO=<epoch> RECON_DLQ=<dlq-count> \
+  RECON_DLQ_SPILL_DIR=<the backfill -spill-dir>
 ```
 
 The sample gate cross-checks `messageId` (full precision), `channelId`,
 `channelType`, `spaceId`, and **`visibles`** between MySQL and the ES doc, so a
 silently dropped ACL (`visibles`) field surfaces as `sample_mismatch>0` and the
-gate exits non-zero. `make run-recon` is **mandatory** before switching the read
+gate exits non-zero. Rows the backfill deliberately routed to the DLQ (bad
+payload / non-numeric id / permanent ES reject — already counted in `-dlq`) are
+*expected* to have no ES doc; both reconcile paths exclude them from the sample
+gate (inline reads its in-memory spill, standalone reads the same spill dir via
+`-dlq-spill-dir` / `RECON_DLQ_SPILL_DIR`) so they never false-fail as
+`sample_missing`. `make run-recon` is **mandatory** before switching the read
 alias to a freshly backfilled index (see `docs/forward-migration-v1.9.md`).
 
 ## Build
--- a/README.md
+++ b/README.md
@@ -89,6 +89,30 @@ independent binary/image distinct from `octo-server`.
 > + 中文 analyzer bootstrap. `/metrics` and lag/backlog instrumentation land in
 > phase 7.
 
+## Backfill & reconciliation (must-run gate)
+
+The phase-6 backfill job (`cmd/backfill`) loads existing `message` shards into
+OpenSearch, bypassing Kafka, and is the **only** path that fills the reader's
+safety fields (`spaceId`/`visibles`/`messageSeq`) from the raw MySQL payload.
+
+A count match alone does not prove correctness — it proves only that the row
+*counts* tie, not that each doc's authz/correctness fields are right. **Always run
+the field-level reconciliation gate after a backfill:**
+
+```sh
+# inline after the backfill run (uses the job's own exact DLQ count as authority):
+go run ./cmd/backfill ... -reconcile -from <epoch> -to <epoch>   # count gate + field-level sample gate
+
+# or standalone, any time:
+make run-recon RECON_FROM=<epoch> RECON_TO=<epoch> RECON_DLQ=<dlq-count>
+```
+
+The sample gate cross-checks `messageId` (full precision), `channelId`,
+`channelType`, `spaceId`, and **`visibles`** between MySQL and the ES doc, so a
+silently dropped ACL (`visibles`) field surfaces as `sample_mismatch>0` and the
+gate exits non-zero. `make run-recon` is **mandatory** before switching the read
+alias to a freshly backfilled index (see `docs/forward-migration-v1.9.md`).
+
 ## Build
 
 ```sh
--- a/cmd/backfill/main.go
+++ b/cmd/backfill/main.go
@@ -11,6 +11,8 @@
 //   - 限速 ≤5k docs/s（默认）；checkpoint 续传（已灌 message_id 高水位，与实时游标物理隔离）。
 //   - 真异常 / ES 永久拒绝落本地 DLQ spill 并精确计数，作为对账门权威输入。
 //   - 配置全走 flag / env。运行结束后建议跑 cmd/reconcile（或 -reconcile 内联对账门）核验。
+//     -reconcile 内联门含 count 对账 **+ 字段级抽样比对**（-recon-sample，默认 200）：
+//     count 只证条数、抽样证 reader 契约关键字段（spaceId/visibles/messageSeq 等）内容一致。
 //
 // 🔴 隔离纪律：本作业对真实环境执行需 Yu / 运维显式 sign-off 后低峰触发，绝不自动跑生产。
 //
@@ -68,6 +70,8 @@ func run() error {
 		doRecon    = flag.Bool("reconcile", false, "after backfill, run the reconcile gate (requires -from/-to)")
 		fromUnix   = flag.Int64("from", 0, "reconcile window start (epoch seconds, inclusive)")
 		toUnix     = flag.Int64("to", 0, "reconcile window end (epoch seconds, inclusive; 0 = now)")
+		sampleN    = flag.Int("recon-sample", envInt("BACKFILL_RECON_SAMPLE", 200), "reconcile field-level sample size (0 disables field sampling, count gate only)")
+		maxDet     = flag.Int("recon-sample-max-details", envInt("BACKFILL_RECON_SAMPLE_MAX_DETAILS", 50), "cap on sample-mismatch detail entries in the report")
 	)
 	flag.Parse()
 
@@ -145,7 +149,7 @@ func run() error {
 	}
 
 	if *doRecon {
-		return reconcile(ctx, db, splitCSV(*esAddrs), *esUser, *esPass, *esIndex, tables, *fromUnix, *toUnix, dlq)
+		return reconcile(ctx, db, splitCSV(*esAddrs), *esUser, *esPass, *esIndex, tables, *fromUnix, *toUnix, *sampleN, *maxDet, dlq)
 	}
 	log.Printf("backfill complete. Run cmd/reconcile (or re-run with -reconcile -from/-to) to gate correctness. Total DLQ count: %d", dlq.Count())
 	return nil
@@ -155,8 +159,14 @@ func run() error {
 // 🔴 用**按窗** DLQ 计数（CountInWindow）：reconcile 窗可能不覆盖本次 run 处理的全部行
 // （如分批跑、或只校验某个时间段），用全量 dlqCount 会把窗外的 DLQ 行也减掉 → false
 // mismatch/false OK（codex review P2-window）。
+//
+// 🔴 字段级抽样门（YUJ-4701 落地 Jerry-Xin non-blocking）：count 门只证「条数」一致，不证
+// 「内容」一致。复用 recon.CompareSamples 逐字段核对 reader 契约关键字段（messageId/channelId/
+// channelType/spaceId/visibles/messageSeq），检出「条数对得上但字段错位」的静默 drift——backfill
+// 正是富化 spaceId/visibles/messageSeq 的唯一路径，这些字段错位最该在 backfill 后立即拦住。
+// sampleN<=0 时退回纯 count 门（与 cmd/reconcile -sample 0 同口径）。
 // 对平退出码 0；不对平返回错误（main 退出码 1）作为 STOP 信号。
-func reconcile(ctx context.Context, db *sql.DB, esAddrs []string, user, pass, index string, tables []string, fromUnix, toUnix int64, dlq *backfill.DLQSpill) error {
+func reconcile(ctx context.Context, db *sql.DB, esAddrs []string, user, pass, index string, tables []string, fromUnix, toUnix int64, sampleN, maxDet int, dlq *backfill.DLQSpill) error {
 	to := toUnix
 	if to == 0 {
 		to = time.Now().Unix()
@@ -200,9 +210,30 @@ func reconcile(ctx context.Context, db *sql.DB, esAddrs []string, user, pass, in
 	if err != nil {
 		return fmt.Errorf("reconcile inputs not self-consistent: %w", err)
 	}
-	log.Printf("reconcile gate: %s", report.String())
-	if !report.OK {
-		return fmt.Errorf("reconcile MISMATCH (diff=%d): backfill correctness gate FAILED — STOP, do not proceed", report.Diff)
+
+	full := recon.FullReport{Count: report, RanAtUnixSeconds: time.Now().Unix()}
+	full.Window.FromUnix = fromUnix
+	full.Window.ToUnix = to
+
+	// 字段级抽样门（count 门不证内容一致）。
+	if sampleN > 0 {
+		sampleReader := recon.NewMySQLSampleReader(db, tables)
+		docFetcher := recon.NewOSDocFetcher(osClient, index)
+		// 排除本批已落 DLQ spill 的真异常 / 永久拒绝行：它们本就不该在 ES 正文索引里
+		// （count 门已用同一 DLQ 计数抵消），抽样命中时不能再算成 missing（口径冲突 → false MISMATCH）。
+		excluded := dlq.MessageIDsInWindow(fromUnix, to)
+		sample, serr := recon.CompareSamplesExcluding(ctx, sampleReader, docFetcher, fromUnix, to, sampleN, maxDet, excluded)
+		if serr != nil {
+			return fmt.Errorf("sample compare: %w", serr)
+		}
+		full.Sample = sample
+	}
+
+	log.Printf("reconcile gate: %s", full.String())
+	if !full.Healthy() {
+		return fmt.Errorf("reconcile MISMATCH (count diff=%d, sample mismatch=%d missing=%d): "+
+			"backfill correctness gate FAILED — STOP, do not proceed",
+			report.Diff, full.Sample.Mismatch, full.Sample.Missing)
 	}
 	return nil
 }
--- a/cmd/reconcile/main.go
+++ b/cmd/reconcile/main.go
@@ -5,10 +5,15 @@
 //
 //	reconcile -from 1709078400 -to 1718323200 \
 //	  -mysql-dsn 'user:pass@tcp(host:3306)/im_prod' -tables message,message1,message2,message3,message4 \
-//	  -es http://localhost:9200 -es-index octo-message [-dlq N]
+//	  -es http://localhost:9200 -es-index octo-message [-dlq N] [-dlq-spill-dir DIR]
 //
 // 与 indexer 写入器解耦（沿用 internal/esindex 解耦纪律）：本命令只读 MySQL + 查 ES，不依赖
 // consumer。阶段 6 backfill job 可直接 import internal/recon 复用对账算术。
+//
+// DLQ 行处理：count 对账用 -dlq 抵消窗内已知 DLQ 数；字段级抽样门用 -dlq-spill-dir 读 backfill
+// 落下的 DLQ spill 文件，把这些行（本就不该有 ES doc）从抽样比对排除——与 inline backfill 对账门
+// （cmd/backfill 用 in-memory DLQSpill.MessageIDsInWindow）口径一致，避免抽样命中合法 DLQ 行时
+// 误报 sample_missing → 误 exit 2 阻塞 alias 切换。
 package main
 
 import (
@@ -25,6 +30,7 @@ import (
 	"strings"
 	"time"
 
+	"github.com/Mininglamp-OSS/octo-search-indexer/internal/backfill"
 	"github.com/Mininglamp-OSS/octo-search-indexer/internal/recon"
 	_ "github.com/go-sql-driver/mysql"
 	"github.com/opensearch-project/opensearch-go/v3"
@@ -50,6 +56,7 @@ func run() error {
 		esUser   = flag.String("es-user", os.Getenv("RECON_ES_USER"), "OpenSearch username")
 		esPass   = flag.String("es-pass", os.Getenv("RECON_ES_PASS"), "OpenSearch password")
 		dlq      = flag.Int64("dlq", 0, "known DLQ count in window (rows that never reached ES body index)")
+		dlqDir   = flag.String("dlq-spill-dir", os.Getenv("RECON_DLQ_SPILL_DIR"), "optional backfill DLQ spill dir (or RECON_DLQ_SPILL_DIR); its message_ids are excluded from the field-level sample gate so legit DLQ rows don't false-fail as sample_missing")
 		sampleN  = flag.Int("sample", envInt("RECON_SAMPLE", 200), "field-level sample size (0 disables sampling)")
 		maxDet   = flag.Int("sample-max-details", envInt("RECON_SAMPLE_MAX_DETAILS", 50), "cap on mismatch detail entries in the report")
 		jsonOut  = flag.Bool("json", false, "emit the structured FullReport as JSON")
@@ -132,7 +139,15 @@ func run() error {
 	if *sampleN > 0 {
 		sampleReader := recon.NewMySQLSampleReader(db, splitCSV(*tablesS))
 		docFetcher := recon.NewOSDocFetcher(osClient, *esIndex)
-		sample, serr := recon.CompareSamples(ctx, sampleReader, docFetcher, *fromUnix, to, *sampleN, *maxDet)
+		// 排除 backfill 已落 DLQ spill 的真异常 / 永久拒绝行：它们本就不该在 ES 正文索引里
+		// （count 门已用同一 DLQ 计数抵消），抽样命中时不能再算成 missing（口径冲突 → false MISMATCH）。
+		// 与 inline backfill 对账门（cmd/backfill）行为一致：那条路用 in-memory DLQSpill.MessageIDsInWindow，
+		// 这条 standalone 路无 live spill，故从 backfill job 留下的 spill 文件只读复原同一份排除集。
+		excluded, lerr := backfill.LoadDLQMessageIDsInWindow(*dlqDir, *fromUnix, to)
+		if lerr != nil {
+			return fmt.Errorf("load DLQ spill exclusion set: %w", lerr)
+		}
+		sample, serr := recon.CompareSamplesExcluding(ctx, sampleReader, docFetcher, *fromUnix, to, *sampleN, *maxDet, excluded)
 		if serr != nil {
 			return serr
 		}
--- a/cmd/reconcile/main.go
+++ b/cmd/reconcile/main.go
@@ -182,7 +182,11 @@ func pushReport(ctx context.Context, url, token string, payload recon.PushPayloa
 	if err != nil {
 		return err
 	}
-	defer func() { _ = resp.Body.Close() }()
+	defer func() {
+		if cerr := resp.Body.Close(); cerr != nil {
+			log.Printf("reconcile: close push response body: %v", cerr)
+		}
+	}()
 	if resp.StatusCode < 200 || resp.StatusCode >= 300 {
 		return fmt.Errorf("push returned status %d", resp.StatusCode)
 	}
--- a/cmd/reconcile/main.go
+++ b/cmd/reconcile/main.go
@@ -12,12 +12,16 @@
 package main
 
 import (
+	"bytes"
 	"context"
 	"database/sql"
+	"encoding/json"
 	"flag"
 	"fmt"
 	"log"
+	"net/http"
 	"os"
+	"strconv"
 	"strings"
 	"time"
 
@@ -46,6 +50,11 @@ func run() error {
 		esUser   = flag.String("es-user", os.Getenv("RECON_ES_USER"), "OpenSearch username")
 		esPass   = flag.String("es-pass", os.Getenv("RECON_ES_PASS"), "OpenSearch password")
 		dlq      = flag.Int64("dlq", 0, "known DLQ count in window (rows that never reached ES body index)")
+		sampleN  = flag.Int("sample", envInt("RECON_SAMPLE", 200), "field-level sample size (0 disables sampling)")
+		maxDet   = flag.Int("sample-max-details", envInt("RECON_SAMPLE_MAX_DETAILS", 50), "cap on mismatch detail entries in the report")
+		jsonOut  = flag.Bool("json", false, "emit the structured FullReport as JSON")
+		pushURL  = flag.String("push-url", os.Getenv("RECON_PUSH_URL"), "optional octo-server ingestion URL to POST the search_recon gauge payload")
+		pushTok  = flag.String("push-token", os.Getenv("RECON_PUSH_TOKEN"), "optional bearer token for -push-url")
 		timeout  = flag.Duration("timeout", 60*time.Second, "overall timeout")
 	)
 	flag.Parse()
@@ -114,12 +123,79 @@ func run() error {
 		// 输入不自洽（DLQ accounting 加固，阶段 6 (f)）：拒绝在可疑计数上判 OK。
 		return err
 	}
-	fmt.Println(report.String())
-	if !report.OK {
-		// 不对平：退出码 2，作为 backfill/CI gate 的失败信号。
+
+	full := recon.FullReport{Count: report, RanAtUnixSeconds: time.Now().Unix()}
+	full.Window.FromUnix = *fromUnix
+	full.Window.ToUnix = to
+
+	// 抽样字段比对（步骤 5 核心：count 对账不证内容一致）。
+	if *sampleN > 0 {
+		sampleReader := recon.NewMySQLSampleReader(db, splitCSV(*tablesS))
+		docFetcher := recon.NewOSDocFetcher(osClient, *esIndex)
+		sample, serr := recon.CompareSamples(ctx, sampleReader, docFetcher, *fromUnix, to, *sampleN, *maxDet)
+		if serr != nil {
+			return serr
+		}
+		full.Sample = sample
+	}
+
+	// 回填 octo-server 只读 search_recon_* gauge（可选）。push 失败不改变对账结论，但要显式告警。
+	if *pushURL != "" {
+		if perr := pushReport(ctx, *pushURL, *pushTok, full.PushPayload()); perr != nil {
+			log.Printf("warn: push recon payload to %s failed: %v", *pushURL, perr)
+		}
+	}
+
+	if *jsonOut {
+		b, jerr := json.MarshalIndent(full, "", "  ")
+		if jerr != nil {
+			return fmt.Errorf("marshal report: %w", jerr)
+		}
+		fmt.Println(string(b))
+	} else {
+		fmt.Println(full.String())
+	}
+
+	if !full.Healthy() {
+		// 不对平（count 或抽样 drift）：退出码 2，作为 backfill/CI/cron gate 的失败信号。
 		os.Exit(2)
 	}
 	return nil
+}
+
+// pushReport POST 结构化 gauge 载荷到 octo-server 只读 ingestion 端（载荷逐字段对齐
+// recon_metrics.go::ReconReport）。仅在 -push-url 配置时调用。
+func pushReport(ctx context.Context, url, token string, payload recon.PushPayload) error {
+	b, err := json.Marshal(payload)
+	if err != nil {
+		return fmt.Errorf("marshal push payload: %w", err)
+	}
+	req, err := http.NewRequestWithContext(ctx, http.MethodPost, url, bytes.NewReader(b))
+	if err != nil {
+		return err
+	}
+	req.Header.Set("Content-Type", "application/json")
+	if token != "" {
+		req.Header.Set("Authorization", "Bearer "+token)
+	}
+	resp, err := http.DefaultClient.Do(req)
+	if err != nil {
+		return err
+	}
+	defer func() { _ = resp.Body.Close() }()
+	if resp.StatusCode < 200 || resp.StatusCode >= 300 {
+		return fmt.Errorf("push returned status %d", resp.StatusCode)
+	}
+	return nil
+}
+
+func envInt(k string, def int) int {
+	if v := os.Getenv(k); v != "" {
+		if n, err := strconv.Atoi(v); err == nil {
+			return n
+		}
+	}
+	return def
 }
 
 func envOr(k, def string) string {
--- a/cmd/reconcile/main_test.go
+++ b/cmd/reconcile/main_test.go
@@ -1,0 +1,104 @@
+package main
+
+import (
+	"context"
+	"fmt"
+	"os"
+	"path/filepath"
+	"testing"
+
+	"github.com/Mininglamp-OSS/octo-search-indexer/internal/backfill"
+	"github.com/Mininglamp-OSS/octo-search-indexer/internal/recon"
+)
+
+// fakeSampleSrc / fakeESFetch 模拟 standalone reconcile 的两个抽样依赖（MySQL 源 + ES doc）。
+type fakeSampleSrc struct{ rows []recon.SampleRow }
+
+func (f *fakeSampleSrc) SampleRows(_ context.Context, _, _ int64, limit int) ([]recon.SampleRow, error) {
+	if limit < len(f.rows) {
+		return f.rows[:limit], nil
+	}
+	return f.rows, nil
+}
+
+type fakeESFetch struct{ docs map[string]recon.ESDocFields }
+
+func (f *fakeESFetch) FetchDocs(_ context.Context, ids []string) (map[string]recon.ESDocFields, error) {
+	out := map[string]recon.ESDocFields{}
+	for _, id := range ids {
+		if d, ok := f.docs[id]; ok {
+			out[id] = d
+		}
+	}
+	return out, nil
+}
+
+// writeSpillWithDLQRow 落一行 DLQ spill 记录（NDJSON + 已 fsync 的 offset sidecar），返回 spill dir。
+// 直接按 backfill spill 的磁盘格式写，模拟 backfill job 跑完后留在盘上、供 standalone reconcile 只读
+// 加载的文件（DLQSpill 的内部写入器不导出，故测试按已固定的盘上契约构造）。
+func writeSpillWithDLQRow(t *testing.T, messageID string, createdAt int64) string {
+	t.Helper()
+	dir := filepath.Join(t.TempDir(), "spill")
+	if err := os.MkdirAll(dir, 0o750); err != nil {
+		t.Fatalf("mkdir spill: %v", err)
+	}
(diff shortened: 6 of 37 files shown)


mochashanyao (approved):
[Octo-Q · automated review]

---

# octo-search-indexer PR#10 Review Report
**PR:** Mininglamp-OSS/octo-search-indexer#10
**Head SHA:** `ac390d0bce43f3ce42d0271c5ed022bda013115d`
**Title:** feat(esindex): converge OS doc contract to reader shape + recon sampling
**Routing:** complexity=security_sensitive (automated review)

---

## 1. Verification Summary

| Item | Status | Evidence |
|---|---|---|
| Tests pass | ✅ | `go test ./...` — all 4 packages OK |
| go vet clean | ✅ | `go vet ./...` — no issues |
| Doc struct aligns to reader | ✅ | `internal/esindex/doc.go:18-42` — camelCase JSON tags match reader `source.go::Doc` |
| messageId long precision | ✅ | `doc.go:82-86` parseMessageID uses int64; `doc_test.go:88-99` round-trips int64 max |
| Non-numeric message_id → DLQ | ✅ | `writer.go:155-160` marks permanent(400); `writer_test.go:276-325` verifies no ES call for poison |
| Consumer safety gate | ✅ | `service.go:100-107` refuses start when SchemaVersion<2; `safetygate_test.go` asserts |
| Backfill enriches spaceId/visibles/messageSeq | ✅ | `doc_enrich.go:22-48`; `doc_enrich_test.go:36-62` |
| Recon sampling field-level | ✅ | `sample.go:82-121` CompareSamples; `sample_test.go` covers match/missing/mismatch/signal |
| Mapping v1.9 correct | ✅ | `octo-message.json` — camelCase fields, nested payload, IK analyzer, alias embedded |
| Forward migration script | ✅ | `scripts/forward-migrate.sh` — 3-step with alias strip + atomic switch + rollback |
| Bulk/BulkDocs index alignment | ✅ | `writer.go:148-186` — docIdx mapping preserves msgs↔out alignment even on batch failure |

## 2. Findings

### P2-1: `Doc.To` field declared but never populated

**Diff-scope:** new (introduced by this PR in `doc.go:25`)

`Doc.To` (`json:"to,omitempty"`) is declared in the struct and mapped in `octo-message.json` as `keyword`, but `DocFromMessage` (`doc.go:88-107`) never sets it. `searchmsg.Message` has no `ToUID` field. The `omitempty` tag means it won't appear in ES docs when empty, so no mapping error. But if the reader ever queries `to` for p2p participant filtering, it'll always be absent.

**Impact:** No runtime error today. Forward-compat concern only — if reader starts relying on `to`, all indexer-written docs will be missing it.

**Recommendation:** Add a `// TODO(phase-N): populate from enriched contract` comment or track as a follow-up.

### P2-2: Painless reindex script fragile on non-numeric `message_id`

**Diff-scope:** new (introduced by this PR in `scripts/forward-migrate.sh:55-68`)

The reindex painless script does `Long.parseLong(ctx._source.remove('message_id'))`. If any doc in the old index has a non-numeric `message_id` (e.g., from old harness seeds like `"m-en"`, `"h-bad"`), the script throws per-doc and the reindex reports `"failures":[...]`, which the script catches via `grep -q '"failures":\[\]'` (exit non-zero). But there's no per-doc skip/retry — the entire reindex halts.

**Impact:** Production should only have numeric snowflake IDs, so low probability. But if any test/staging data leaked into the old index, migration blocks with no graceful degradation.

**Recommendation:** Add a painless guard: `if (ctx._source.message_id instanceof String && !ctx._source.message_id.matches('\\d+')) { ctx.op = 'noop'; }` before the parse.

### P2-3: `parsePayloadVisibility` duplicated between `recon/sample.go` and `backfill/doc_enrich.go`

**Diff-scope:** new (both functions introduced by this PR)

`recon/sample.go:253-271` (`parsePayloadVisibility`) and `backfill/doc_enrich.go:57-81` (`extractVisibility`) implement the same JSON parsing logic for `space_id` + `visibles`. Both filter non-string elements identically. If one is updated without the other, recon and backfill will disagree on visibility extraction.

**Impact:** Maintainability risk. No current bug.

**Recommendation:** Extract to a shared package (e.g., `internal/visibility`) or at minimum add a cross-reference comment.

### P2-4: `PushPayload` conflates `Mismatch` and `Missing` into single gauge

**Diff-scope:** new (introduced by this PR in `report.go:49-55`)

`PushPayload.SampleMismatch = int64(r.Sample.Mismatch + r.Sample.Missing)` merges two distinct failure modes (field drift vs. absent docs) into one metric. The `FullReport.Healthy()` method checks them separately (`r.Sample.Mismatch == 0 && r.Sample.Missing == 0`), but the pushed gauge loses this distinction.

**Impact:** Operators seeing `search_recon_sample_mismatch > 0` can't tell if it's field drift (more urgent) or missing docs (may overlap with `doc_drift`).

**Recommendation:** Push two separate gauges, or add a `sample_missing` field to `PushPayload`.

### Nit-1: Double assignment in `docFromRow` for negative `MessageSeq`

`doc_enrich.go:41-44`:
```go
doc.MessageSeq = uint64(row.MessageSeq)  // if negative → large uint64
if row.MessageSeq < 0 {
    doc.MessageSeq = 0  // immediately overwritten
}
```
Functionally correct (negative is clamped to 0 before any consumer sees it), but the intermediate large-uint64 assignment is dead code. Minor readability nit.

### Nit-2: Recon sample ordering by VARCHAR `message_id`

`sample.go:213`: `ORDER BY message_id ASC` on a VARCHAR(20) column gives lexicographic ordering ("9" > "100"). Deterministic and reproducible (as documented), but doesn't sample the chronologically earliest messages. Consider `ORDER BY id ASC` for temporal coverage.

## 3. Recommendations

1. **P2-2 is the most actionable before merge** — add a painless guard for non-numeric `message_id` in the reindex script to prevent migration halt on stale test data.
2. P2-3 and P2-4 can be tracked as follow-ups.
3. The overall architecture is sound: dual-path (backfill vs consumer) convergence through `BulkDocs`, safety gate on consumer start, and field-level recon sampling form a coherent defense-in-depth strategy.

## 4. Extra Findings

- **Mapping `dynamic: strict`** is correctly preserved — any doc with unexpected fields will be rejected by ES, preventing silent index pollution.
- **Alias staging** in the migration script (strip aliases on index creation, add atomically in step 3) correctly prevents the read alias from pointing to a half-populated index.
- **DLQ `Reason` field** is properly set in the new `docFromRow` path (`"payload_parse"` vs `"doc_convert"`) and falls back to default in `DLQSpill.Write` — no accounting gap.

## 5. Data-Flow Tracing

|---|---|---|
| `Doc.MessageID` (int64) | `parseMessageID(msg.MessageID)` → `strconv.ParseInt` | ✅ Non-numeric → error → DLQ(400). Never silently 0. |
| `Doc.MessageSeq` (uint64) | backfill: `srcMessageRow.MessageSeq` (SQL `message_seq BIGINT`); consumer: not in Kafka contract → 0 | ✅ Backfill clamps negative to 0. Consumer path blocked by safety gate. |
| `Doc.SpaceID` (string) | backfill: `extractVisibility(payload)` → `payload.space_id`; consumer: empty | ✅ Backfill extracts from MySQL payload. Consumer blocked. Signal encrypted → empty (fail-closed). |
| `Doc.Visibles` ([]string) | backfill: `extractVisibility(payload)` → `payload.visibles`; consumer: empty | ✅ Non-string elements filtered. Consumer blocked by safety gate (prevents fail-OPEN). |
| `Doc.Payload.Text.Content` | `buildPayload(contentType, content)` → only when type==Text && content non-nil/non-empty | ✅ Non-text → `payload.text=nil`. Reader handles absent text gracefully. |
| `BulkItemResult.MessageID` | backfill: `Doc.idString()` = `FormatInt(MessageID, 10)`; consumer: `msgs[i].MessageID` | ✅ Both paths produce consistent string IDs. Backfill/consumer write same `_id` → idempotent. |
| Recon `SampleRow.MessageSeq` | SQL `message_seq` column via `scanSampleRows` | ✅ Compared against `ESDocFields.MessageSeq` (uint64) via string formatting. Type mismatch for negatives detected as drift (correct). |
| Recon `ESDocFields` | `OSDocFetcher.FetchDocs` → `json.Unmarshal(resp.Hits.Hits[i].Source)` | ✅ Uses reader-aligned JSON tags (`messageId`, `channelId`, etc.). Field name typo would surface as mismatch in recon. |
| `FullReport.Healthy()` | `Count.OK && Sample.Mismatch==0 && Sample.Missing==0` | ✅ When sampling disabled (`sampleN=0`), Sample is zero-value → Healthy reduces to Count.OK. Correct. |

## 6. R5 Blind-Spot Checklist (security_sensitive)

**C1 — Dual-path parity:**
- backfill (`BulkDocs(docs)`) and consumer (`Bulk(msgs)` → `DocFromMessage` → `BulkDocs(docs)`) converge at the same write path with identical `_id` encoding (`idString()` = `FormatInt(MessageID, 10)`). ✅ Clear.
- The asymmetry (backfill enriches, consumer doesn't) is **intentional and safety-critical**: consumer is blocked by `LiveContractCarriesSafetyFields()` until the Kafka contract carries safety fields. ✅ Clear.

**C2 — Control-flow ordering / nesting reuse:**
- `encodeBulkBody` is called from `BulkDocs` only (not `Bulk` directly). `Bulk` delegates to `BulkDocs` after conversion. No double-encoding or ordering issue. ✅ Clear.
- No security controls (regex/escape/sanitize) in this PR. N/A.

**C3 — Authorization boundary ≠ capability boundary:**
- This is an indexer writing to ES. No new endpoints, tools, or user-facing capabilities are exposed. The `pushReport` HTTP POST is gated by `-push-url` flag and uses bearer token auth. ✅ Clear.

---

**[Octo-Q] verdict: APPROVE**

No P0/P1 findings. The PR correctly converges the indexer's OS doc contract to the reader shape, adds a well-designed consumer safety gate (fail-closed on missing Kafka contract fields), implements field-level recon sampling, and provides a controlled forward migration script. All P2 findings are maintainability/robustness improvements that don't block landing.

Jerry-Xin (dismissed):
The PR is in scope for `octo-search-indexer`, but it currently breaks the documented Kafka end-to-end harness.

**🔴 Blocking**
- 🔴 Critical: The live-ingestion safety gate makes the existing Kafka harness unusable. `Service.Run` now immediately returns while `searchmsg.SchemaVersion` is below `SafetyFieldsSchemaVersion` (`internal/consumer/service.go:103`, `internal/esindex/doc.go:125`), and the current vendored contract is still version 1. However `harness/run.sh` still starts `go run ./cmd/es-indexer` as the core of the local E2E (`harness/run.sh:87`) and the README still documents `./harness/run.sh` as the full Kafka → consumer → OpenSearch verification path (`harness/README.md:31`). In this PR state the indexer exits before consuming, then seed/verify cannot pass. Fix by updating/removing this harness path for v1.9, or by providing a strictly local/test-only v2 contract fixture path that does not weaken the production safety gate.

**💬 Non-blocking**
- 🟡 Warning: `cmd/backfill -reconcile` still runs only the count gate, not the new field-level sample gate (`cmd/backfill/main.go:147`, `cmd/backfill/main.go:194`). Operators using the inline backfill gate can still miss the exact field drift this PR is adding `cmd/reconcile` to catch. Consider reusing `FullReport`/`CompareSamples` there or documenting that `make run-recon` is now mandatory after backfill.
- 🔵 Suggestion: The rollback snippet printed by `scripts/forward-migrate.sh` is emitted from a single-quoted heredoc, so `$ES`, `$NEW_INDEX`, `$ALIAS`, and `$OLD_INDEX` remain literal despite the embedded shell-quote construction (`scripts/forward-migrate.sh:85`). Either make it a true template or expand it into a ready-to-run command.

**✅ Highlights**
- The writer correctly treats non-numeric `message_id` values as permanent per-item failures instead of coercing to `0`.
- Backfill enrichment is scoped well: `spaceId`, `visibles`, and `messageSeq` are filled only where the MySQL source can provide them.
- Reconciliation now checks `_source.messageId` precision, which is important for snowflake IDs and cursor correctness.
- `go test ./...` passes locally.

Jerry-Xin:
**Addendum from reviewer (gate + severity calibration)**

The REQUEST_CHANGES above stands, but the **decisive blocker is CI, not the harness**. Calibrating:

**🔴 Blocking (CI Lint is red on this PR's own new code — must be green before merge)**
golangci-lint reports 4 new issues introduced by this diff:
- `cmd/reconcile/main.go:185` — errcheck: `resp.Body.Close` return value not checked
- `internal/recon/sample.go:222` — errcheck: `rows.Close` return value not checked
- `internal/esindex/doc_test.go:86` — errcheck: `json.Marshal` return value not checked
- `internal/backfill/runner.go:151` — staticcheck QF1002: could use tagged switch

These are quick fixes but the Lint gate is required, so the PR cannot merge as-is.

**🟡 Re-severity of the harness finding (was flagged 🔴 above → it's actually a follow-up):**
The fail-closed gate in `internal/consumer/service.go:103` (refuse live ingestion while `searchmsg.SchemaVersion(1) < SafetyFieldsSchemaVersion(2)`) is **correct, intentional safety design** — it prevents the reader's empty-`visibles` fail-OPEN leak. That is a highlight, not a defect. Its side effect is that the *live-Kafka* path of `harness/run.sh` (`go run ./cmd/es-indexer`) now exits immediately; `run.sh`/README weren't updated to note this. That's dev-harness staleness (the v1.9-relevant `harness/backfill` E2E path was correctly updated), so treat it as 🟡 follow-up, not a production blocker.

**✅ Verified correct (the high-risk patterns for this repo are handled well):**
- OS-doc-contract convergence matches reader shape byte-for-byte: camelCase nested doc (`messageId` long, `channelId`, `channelType`, `spaceId`, `visibles`, `messageSeq`, `payload.text.content` IK). 
- Snowflake precision: reconcile reads typed `MessageID int64` from `_source.messageId` (`internal/recon/sample.go:61,121-124`), NOT the float64 Sort — cursor/tiebreaker stays precise. 
- Visibility gating: the safety-fields gate is the right fail-closed posture; backfill enriches `spaceId`/`visibles`/`messageSeq` from MySQL.

**Action for author:** fix the 4 lint issues to get CI Lint green (also `Closes #<issue>` for check-sprint), and either update/annotate `harness/run.sh`+README for the gated live path or land that as a quick follow-up. Re-request review once Lint is green.

yujiawei (commented):
# Code Review — PR #10 (octo-search-indexer)

Reviewed at head `ac390d0bce43f3ce42d0271c5ed022bda013115d` against merge-base `c79b9242` (base `main`). 28 files, +1721/-183. This review independently re-derived the diff and merge-base; cached review signals were not trusted.

Local gates I ran on this head: `go build ./...` ✅, `go vet ./...` ✅, `go test ./...` ✅ (all packages pass). I did not separately re-run `-race`; the PR claims it passes.

This is a well-disciplined change overall — fail-closed live gate, durability ordering before checkpoint advance, TOCTOU-safe `EnsureIndex`, recon shard-failure guards, and full-precision `messageId`. Two blocking issues remain, both with low-cost fixes, both security-relevant in a visibility-gating change, so I'm flagging them now rather than letting the phase-9 operator step on them.

## 1. Verification summary

- ✅ Doc contract converges to the reader shape: camelCase + nested `payload.*`, `messageId` as `long` (full snowflake precision, `idString()` via `FormatInt`), sort fields `timestamp`+`messageId`. Verified in `internal/esindex/doc.go`, `mapping/octo-message.json`.
- ✅ Non-numeric `message_id` is a permanent (DLQ) error, never coerced to `0` — `parseMessageID` (`doc.go:62`), `Bulk` per-item `Status=400` (`writer.go:155`), backfill `docFromRow` → `outcomeDLQ` (`doc_enrich.go:28`).
- ✅ Backfill self-sources `spaceId`/`visibles` from raw payload and `messageSeq` from the column; encrypted (Signal) rows skip payload parsing → empty (reader fail-closed). `internal/backfill/doc_enrich.go`, `extract.go`, `source.go`.
- ✅ Live-ingestion gate refuses to start while the Kafka contract lacks safety fields. Verified `consumer/service.go:Run` checks the gate **before** `EnsureIndex`, and `searchmsg.Message` (octo-lib) genuinely lacks `SpaceID`/`Visibles`/`MessageSeq`, so the gate is closed today. ✅ — **but see P1-2 for the latent trap when it opens.**
- ✅ Recon count + sample gate: shard-completeness guards on both `_count` and `_search` (`collectors.go:130`, `sample.go:FetchDocs`), input self-consistency validation (`recon.go:Validate`), exit-2 on unhealthy.
- ⚠️ Migration runbook + embedded alias — **P1-1**.

## 2. Findings (P1 — blocking)

### P1-1 — Embedded read alias is attached on index creation, defeating the "no mixed old/new docs under one alias" invariant
`internal/esindex/mapping/octo-message.json:7` (`aliases.wukongim-messages-read`) + `internal/esindex/ensure_index.go:55` (`Indices.Create` sends the embedded mapping verbatim) + `docs/forward-migration-v1.9.md` ("生产推荐：直接 backfill 重灌新索引").

`forward-migrate.sh` STEP=1 deliberately strips the `aliases` block (`jq 'del(.aliases)'`) precisely so the read alias is not attached before reconciliation passes. But the runbook's *recommended* alternative — run `cmd/backfill` straight into a new `BACKFILL_ES_INDEX` — goes through `backfill.Run → EnsureIndex → Indices.Create`, which sends the **un-stripped** embedded mapping and therefore attaches `wukongim-messages-read` to the brand-new, still-filling index. Since an alias may point at multiple indices, the read alias then fans out over both the legacy flat index (unreadable by the new reader) and the partially-populated new index → mixed-contract / garbled reads during migration. This directly contradicts the invariant the script and docs both assert ("禁半新半旧 doc 同 alias 服务").

Why it matters: live read corruption during the migration window, triggered by following the shipped runbook as written. Low-cost fixes (any one): strip the alias on the programmatic `EnsureIndex` create path too; or remove the alias from the embedded mapping and attach it only via the staged `_aliases` step; or document explicitly that backfill must target an index pre-created (alias-stripped) by STEP=1 and must not rely on `EnsureIndex` to create the new index.

### P1-2 — Live-ingestion safety gate keys only on schema version; `DocFromMessage` never populates the safety fields (latent fail-open)
`internal/esindex/doc.go:86` (`SafetyFieldsSchemaVersion = 2`), `doc.go:129` (`LiveContractCarriesSafetyFields`), `doc.go:DocFromMessage` (does not set `SpaceID`/`Visibles`/`MessageSeq`), `consumer/service.go:Run`.

The gate's contract is "`searchmsg.SchemaVersion >= 2` ⇒ safe to write live." That condition is **necessary but not sufficient**: when octo-lib is bumped to SchemaVersion 2 (phase 9), the gate opens, but nothing guarantees `DocFromMessage` has been updated to copy the three safety fields. The fields don't even exist on `searchmsg.Message` yet, and no test or compile-time link couples "gate open" to "DocFromMessage populates safety fields." A phase-9 implementer who bumps the version and the producer but forgets to wire `DocFromMessage` silently reintroduces the exact reader fail-open (ordinary members searching admin-only group/system messages) this gate exists to prevent.

Since the gate-open logic ships in *this* PR, the invariant should be pinned here. Low-cost fix: a guard test that fails if `LiveContractCarriesSafetyFields()` is true while `DocFromMessage` leaves `Visibles`/`SpaceID`/`MessageSeq` empty for a message carrying them; or at minimum a loud `DocFromMessage` comment/TODO tying the two together. (Today the gate is closed, so there is no active leak — this is strictly a latent trap, but a security-relevant one.)

## 3. Findings (P2 / nits — non-blocking)

- **P2-1 — Recon sample path has no DLQ accounting** (`internal/recon/sample.go:CompareSamples`, `report.go:Healthy`). Count recon subtracts known DLQ via `-dlq`, but sample comparison counts any sampled MySQL row absent from ES as `Missing` → unhealthy → exit 2. A sampled, legitimately-DLQ'd row (poison pill / parse failure) therefore yields a false MISMATCH. Consider excluding known-DLQ ids from the sample population, or tolerating sampled rows that map to DLQ records.

- **P2-2 — `make test` masks failing tests** (`Makefile:18`). `$(GO) test ./... | tail -40` makes the recipe's exit status come from `tail`, not `go test` (default `sh`, no `pipefail`). Reproduced locally: `false | tail -1` → exit 0. Failing tests can show as a green `make test`. CI appears to call `go test` directly so this is a local-gate footgun, but worth fixing (`.SHELLFLAGS = -ec -o pipefail` or drop the pipe).

- **P2-3 — `RECON_ES_INDEX` defaults to legacy `octo-message` while recon expects v1.9 camelCase** (`cmd/reconcile/main.go:53`, `Makefile:28`, vs `recon/collectors.go:rangeFilter` + `sample.go:ESDocFields`). Running `make run-recon` with defaults against the still-legacy index queries `createdAt`/`messageId`/`spaceId` against snake_case docs → false ~100% drift, exit 2. I confirmed recon is **not** wired into this repo's CI/cron (no `run-recon`/`reconcile` reference under `.github/workflows`), so the "breaks CI on merge" impact is overstated — but it is a real operator footgun during transition. Suggest defaulting to the read alias or requiring the flag.

- **P2-4 — Non-sargable recon sample query** (`internal/recon/sample.go:SampleRows`). `WHERE UNIX_TIMESTAMP(created_at) BETWEEN ? AND ?` defeats any index on `created_at`, forcing per-shard full scans each run. Offline tool, matches the pre-existing `CountRows` pattern, so impact is bounded — but `created_at BETWEEN FROM_UNIXTIME(?) AND FROM_UNIXTIME(?)` would let MySQL use the index.

- **P2-5 — Reindex painless has no exception handling** (`scripts/forward-migrate.sh:52`). `Long.parseLong(ctx._source.remove('message_id'))` aborts the whole `_reindex` batch on a single non-numeric legacy `message_id` (legacy mapping typed `message_id` as `keyword`, so non-numeric values are possible). Controlled ops script with loud failure, but a `try/catch` that routes bad ids aside would prevent a mid-migration stall.

- **Nit — `uint64` cast precedes the bounds check** (`internal/backfill/doc_enrich.go:34-37`). `doc.MessageSeq = uint64(row.MessageSeq)` then `if row.MessageSeq < 0 { doc.MessageSeq = 0 }`. Correct as written (the negative case is zeroed before use), but ordering the `< 0` guard first reads more safely.

## 4. Coverage notes / blind spots

- **Cross-repo reader contract not independently verified.** I confirmed `searchmsg.Message` (octo-lib) lacks `SpaceID`/`Visibles`/`MessageSeq` (supports P1-2), but did **not** verify octo-server `modules/messages_search/source.go::Doc` field-by-field — it's a different repo not checked out here. The field-name/type alignment claims (incl. the `recon_metrics.go::ReconReport` push payload shape) rest on the PR's contract table, not on reader source.
- **OS cluster assumptions unverified.** The mapping requires the IK analyzer plugin (`ik_max_word`/`ik_smart`); if the managed cluster lacks it, `EnsureIndex`'s `Create` fails at runtime. Current alias topology (whether `wukongim-messages-read` already points at `octo-message`) is also out of band — relevant to P1-1's blast radius.
- **Painless reindex not executed** against real legacy docs.
- **`Doc.to` is never populated** by either write path (omitempty). Whether the reader needs `to` for p2p recall is a reader-side question (cross-repo, unverified).
- Both advisory passes completed this round (one cross-family long-context pass + one line-level adversarial pass); the findings above are the fused, independently re-verified set. No advisory leg was absent.

## Verdict

**CHANGES_REQUESTED** — on P1-1 (alias attached pre-recon via the recommended backfill path, breaking the PR's own no-mixed-alias invariant) and P1-2 (gate-open invariant is necessary-not-sufficient; bumping the schema version opens live ingestion without guaranteeing the safety fields are written). Both have small fixes and both are security-relevant in a visibility-gating change. The P2/nits are non-blocking.

Jerry-Xin (dismissed):
Summary: The PR is relevant to `octo-search-indexer`, but the live-ingestion safety gate can automatically open before the writer actually emits the required safety fields.

🔴 Blocking

🔴 Critical — The safety gate is keyed only to `searchmsg.SchemaVersion`, but `DocFromMessage` still never copies `spaceId`, `visibles`, or `messageSeq` into the OpenSearch doc. In `internal/consumer/service.go:103-110`, live ingestion is allowed once `LiveContractCarriesSafetyFields()` returns true; that function only checks `searchmsg.SchemaVersion >= 2` in `internal/esindex/doc.go:127-130`. However, `DocFromMessage` at `internal/esindex/doc.go:81-98` still leaves those fields empty. When `octo-lib` bumps `SchemaVersion` to 2, this service will start writing live docs again while still omitting the same fields the gate was meant to protect, including `visibles`, causing the documented reader fail-open risk.  
Fix by either keeping the gate closed until this indexer explicitly maps the new contract fields, or implementing the field mapping and tests as part of the same change that can unlock the gate.

💬 Non-blocking

🟡 Warning — `scripts/forward-migrate.sh:58-67` removes `$ALIAS` from `$OLD_INDEX` unconditionally. If the old flat index never had `wukongim-messages-read`, the atomic alias switch can fail at the final step. Consider checking alias presence first or using a removal form that tolerates absence.

✅ Highlights

The backfill path correctly separates already-enriched `Doc` writes via `BulkDocs`, and the new reconciliation sampler explicitly checks `_source.messageId` precision, `spaceId`, `visibles`, and `messageSeq`.

Verification: `go test ./...` passes locally.

mochashanyao (approved):
[Octo-Q · automated review]

---

# octo-search-indexer PR #10 — automated review Review Report
**Head SHA:** `47186489909691d87ce01da89a3893fc56d995ae`
**Repo:** Mininglamp-OSS/octo-search-indexer
**Scope:** 33 files, +2156/-198 lines — ES doc contract convergence to reader shape + recon sampling

---

## 1. Verification Summary

| Area | Status | Evidence |
|------|--------|----------|
| Doc struct ↔ mapping alignment | ✅ | `doc.go` fields match `octo-message.json` v1.9 (camelCase, nested payload, messageId=long) |
| Safety gate (fail-closed) | ✅ | `service.go:Run()` refuses start when `SchemaVersion < 2`; `LiveContractCarriesSafetyFields()` checks octo-lib constant |
| Backfill enrichment (spaceId/visibles/messageSeq) | ✅ | `doc_enrich.go:docFromRow()` extracts from raw MySQL payload JSON, sets on Doc before `BulkDocs` |
| Non-numeric messageId rejection | ✅ | `doc.go:parseMessageID()` rejects non-numeric → `writer.go:Bulk()` marks Status=400 permanent → DLQ |
| Idempotent upsert | ✅ | `encodeBulkBody()` uses `doc.idString()` (= strconv.FormatInt(messageId,10)) as ES `_id`; index action = upsert |
| Recon count + sample gates | ✅ | `cmd/backfill/main.go` and `cmd/reconcile/main.go` both run `CompareSamplesExcluding` with DLQ exclusion set |
| Forward migration script | ✅ | `scripts/forward-migrate.sh` 3-step: create new index → reindex with painless → atomic alias switch; includes rollback |
| Harness contract gate | ✅ | `harness/contractgate/main.go` probe exits 10 when gated; `run.sh` skips live e2e gracefully |
| Snowflake precision | ✅ | messageId typed `int64` (Go) → `long` (ES mapping); never passes through float64 |
| DLQ dedup + window exclusion | ✅ | `dlq.go:MessageIDsInWindow()` returns non-empty message_ids only; recon excludes these from sample set |

## 2. Findings

### F1 — P2: Bulk() batch-level error overwrites permanent conversion failures (new)

**Diff-scope:** New — introduced by this PR's `Bulk()` refactor that adds `DocFromMessage` conversion before `BulkDocs`.

**Path:** `internal/esindex/writer.go`, `Bulk()` method.

When `DocFromMessage` fails (non-numeric messageId), the entry is correctly marked `Status=400` (permanent) in the output slice. However, if `BulkDocs()` subsequently returns a batch-level error (ES unreachable), the error-handling loop overwrites **all** `docIdx` entries — including the already-permanent ones — with `Status=0` (transient):

```go
// writer.go, Bulk() batch-error path:
for j, idx := range docIdx {
    if j < len(docRes) {
        out[idx] = docRes[j]  // overwrites Status=400 → Status=0
    } else {
        out[idx] = BulkItemResult{OK: false, Status: 0, Err: err}
    }
}
```

**Impact:** Non-convertible entries are retried unnecessarily during transient batch failures. On the next `resolvePass`, `DocFromMessage` fails again → Status=400 again → eventually routes to DLQ when the batch succeeds. **Correctness is preserved** — the entries do reach DLQ eventually — but they cause redundant retry cycles.

**Severity rationale:** P2, not P1. No data loss, no incorrect routing, no user-visible error. The entries reach DLQ after transient errors clear. The fix is simple: skip entries already marked permanent in the batch-error overwrite loop.

**Suggested fix:**
```go
for j, idx := range docIdx {
    if out[idx].Status == docConvertBadRequest {
        continue // preserve permanent conversion failure
    }
    // ... existing overwrite logic
}
```

### F2 — P2: Painless migration script Long.parseLong on potentially non-numeric legacy _id (new)

**Diff-scope:** New — `scripts/forward-migrate.sh` painless reindex script.

The painless script uses `Long.parseLong(ctx._source.message_id)` to convert the legacy `message_id` keyword field to the new `messageId` long field. If any legacy document has a non-numeric `message_id` as its ES `_id`, the reindex will fail for that document.

**Impact:** Migration could fail on edge-case documents. The new code correctly rejects non-numeric IDs (routes to DLQ), but the migration script doesn't handle them.

**Severity rationale:** P2. This is a one-time migration script, not runtime code. The operator can pre-scan for non-numeric `_id` values and exclude them. Adding a try/catch in the painless script would be a nice robustness improvement.

### F3 — P2: Visibility extraction logic duplicated between backfill and recon (new)

**Diff-scope:** New — both `backfill/doc_enrich.go:extractVisibility()` and `recon/sample.go:extractVisibility()` implement the same JSON parse + spaceId/visibles extraction from raw payload bytes.

**Impact:** If one is updated without the other, they'll drift, causing recon to report false mismatches (or miss real ones). The `signalSettingMask` constant is already duplicated in `extract.go` with a comment acknowledging the trade-off.

**Severity rationale:** P2. Maintainability risk, not a current bug. Consider extracting to a shared `internal/visibility` package.

### F4 — P2: PushPayload.SampleMismatch is boolean, loses failure mode detail (new)

**Diff-scope:** New — `recon/report.go:PushPayload()`.

`PushPayload.SampleMismatch` is a single boolean. The JSON report (`SampleResult`) correctly breaks out `Mismatch`, `Missing`, `Skipped`, `Details`. But the push payload (for octo-server gauge ingestion) collapses all sample failures into one bool.

**Impact:** Dashboard/alerting on the gauge can't distinguish "1 field mismatch" from "50 missing docs". The exit-code path (exit 2) has the same limitation.

**Severity rationale:** P2. Observability gap, not a correctness issue. Consider adding `SampleMismatchCount` and `SampleMissingCount` integer fields to `PushPayload`.

### F5 — Note: Doc.To and Doc.Revoked fields defined but never populated (new)

**Diff-scope:** New — `doc.go` defines `To string` and `Revoked bool` fields; mapping includes `to` (keyword) and `revoked` (boolean).

Neither `DocFromMessage()` nor `docFromRow()` sets these fields. They serialize as zero-values (empty string / false) or are omitted via `omitempty`.

**Impact:** None currently. These are forward-compatible placeholders for future producer enrichment. No reader depends on them yet.

**Severity rationale:** Note only. Document the intent or remove until needed.

## 3. Suggestions

1. **F1 fix** (recommended): Add `if out[idx].Status == docConvertBadRequest { continue }` guard in `Bulk()` batch-error overwrite loop. Low-cost, eliminates unnecessary retries.
2. **F2 fix**: Add painless try/catch around `Long.parseLong` with a sentinel value (e.g., -1) for non-numeric IDs, or pre-filter the reindex query.
3. **F3 fix**: Extract shared visibility parsing to `internal/visibility/extract.go` with a single `extractVisibility(payload []byte) (spaceID string, visibles []string)` function.
4. **F4 fix**: Add integer counts to `PushPayload` for richer alerting.

## 4. Additional Observations

- **Harness seed data migration** (harness/seed, harness/verify, harness/backfill): Message IDs changed from string labels ("m-en-1") to numeric snowflake strings ("1000000000000000001"). Verify queries updated for new field names. This is correct and consistent.
- **IK analyzer configuration**: Mapping uses `ik_max_word` for indexing (fine-grained segmentation) and `ik_smart` for search (coarser). This is the standard IK plugin pattern.
- **Alias `wukongim-messages-read`**: Added to the mapping for reader-side decoupling. Good forward-compatibility.
- **`safeTableName` validation** in `recon/sample.go`: Rejects table names with non-alphanumeric/underscore chars. Prevents SQL injection in the sharded table query. Good.

## 5. Data Flow Tracing

### Backfill path (MySQL → ES)
```
srcMessageRow (SQL: id, message_id, message_seq, from_uid, channel_id, channel_type, setting, signal, timestamp, created_unix, payload)
  → extractMessage(row) → searchmsg.Message (SchemaVersion, MessageID, ChannelID, ChannelType, FromUID, MsgTimestamp, CreatedAt, Source, Content/RawExcluded)
  → DocFromMessage(msg) → Doc (messageId=int64 via parseMessageID, channelId, channelType, createdAt, payload.text.content or rawExcluded)
  → docFromRow enrichment:
    - doc.MessageSeq = row.MessageSeq (from SQL message_seq column) ✅
    - extractVisibility(row.Payload) → spaceID, visibles (from payload JSON "space_id" and "visibles" keys) ✅
    - Signal encrypted → skip visibility extraction (spaceId="", visibles=nil) ✅
  → writer.BulkDocs(ctx, docs) → encodeBulkBody → _bulk NDJSON (index action, _id=doc.idString())
  → ES upsert (idempotent)
```

### Live consumer path (Kafka → ES)
```
Kafka → searchmsg.Message → writer.Bulk(ctx, msgs)
  → DocFromMessage(msg) → Doc (same as above)
  → Non-numeric messageId → Status=400 permanent → DLQ ✅
  → BulkDocs(ctx, docs) → _bulk NDJSON → ES upsert
  → Safety gate: service.go:Run() checks LiveContractCarriesSafetyFields() BEFORE EnsureIndex
    → SchemaVersion < 2 → refuse start (fail-closed) ✅
    → spaceId/visibles/messageSeq NOT set on live docs (zero-values) — gated until producer enriches
```

### Recon sampling path
```
MySQL sample (message_id ASC, deterministic) → srcMessageRow
  → extractMessage → searchmsg.Message → DocFromMessage → expected Doc
  → extractVisibility(row.Payload) → expected spaceId/visibles (same logic as backfill) ✅
  → Signal encrypted → skip visibility comparison ✅
OS fetch (ids query → batch get) → actual Doc
  → Compare field-by-field: messageId, messageSeq, channelId, channelType, spaceId, visibles
  → DLQ exclusion: MessageIDsInWindow() returns non-empty message_ids only ✅
```

### Offset commit path (consumer)
```
Bulk results → classifyBulk():
  - Status 2xx → dispCommitted (advance offset)
  - Status 4xx (incl. 400 from docConvertBadRequest) → dispDLQResolved (permanent, advance offset)
  - Status 5xx/429/0 → dispTransient (retry, block offset)
partitionCommitPoints(): walks continuous success prefix per partition
  → offset never advances past unacknowledged transient items ✅
```

## 6. R5 Blind Spot Checklist (security_sensitive PR)

### C1 — Dual-path parity
**Clear.** Verified:
- `Bulk()` (live) and `BulkDocs()` (backfill) share the same `encodeBulkBody` + `mapBulkResults` code path.
- Both paths use `doc.idString()` (= `strconv.FormatInt(d.MessageID, 10)`) as ES `_id` → idempotent overlap.
- `extractVisibility()` is duplicated (F3) but currently identical in both locations.
- No add/remove or subscribe/unsubscribe symmetry pairs in this PR.

### C2 — Control-flow ordering / nested reuse
**Clear.** Verified:
- Safety gate (`LiveContractCarriesSafetyFields()`) runs BEFORE `EnsureIndex()` in `service.go:Run()`. Correct ordering.
- `DocFromMessage()` is called once per message in both paths. No double-conversion.
- `encodeBulkBody()` serializes once per batch. No nested/duplicate encoding.
- `extractVisibility()` JSON-parses the raw payload once. No double-parse.
- No regex/sanitize/escape functions requiring non-canonical form testing.

### C3 — Authorization boundary ≠ capability boundary
**Clear.** Verified:
- This PR does not add new endpoints, tools, or API routes.
- The `visibles` field enrichment is a data-flow change, not an authorization change.
- The safety gate is a **restriction** (fail-closed), not an expansion of capability.
- The `visibles` field in the backfill path is extracted from the message's own payload (the sender's visibility intent), not from the indexer's own permissions. This is correct — the indexer is a data pipeline, not an actor.
- No new Kafka topics or consumer groups are introduced.

---

## Verdict

No P0 or P1 findings. All findings are P2 (efficiency / maintainability / robustness improvements). The PR correctly implements:
- Reader-aligned doc contract with snowflake-safe messageId (long, not string)
- Fail-closed safety gate preventing live ingestion until Kafka contract carries security fields
- Backfill enrichment from raw MySQL payload (spaceId/visibles/messageSeq)
- Recon sampling with DLQ exclusion and field-level comparison
- Forward migration with atomic alias switch and rollback procedure
- Idempotent upsert across live + backfill paths (same `_id`)

[Octo-Q] verdict: **APPROVE**

yujiawei (commented):
# Code Review — PR #10 (octo-search-indexer)

**Head reviewed:** `47186489909691d87ce01da89a3893fc56d995ae` · base `main` (merge-base `c79b9242`)
**Scope:** OS doc contract converged to the reader shape (camelCase nested, `messageId` as `long`, adds `spaceId`/`visibles`/`messageSeq`) + field-level reconciliation sampling gate + forward-migration script. Build / vet / `test -race` all green locally.

This is a security-sensitive change: `visibles` is the access-control allowlist for group system messages, `spaceId` enforces DM space isolation, and `messageSeq` gates cleared-conversation visibility. I verified the consuming side directly against the reader (`octo-server modules/messages_search/visibility.go:216`, `dsl.go:80`): an **empty `visibles` is treated as "no gate" (fail-open)**, and a missing `spaceId` on p2p docs is fail-closed. That asymmetry is what makes the findings below load-bearing.

Overall the engineering quality is high — the DLQ crash-recovery (offset sidecar + truncate-dirty-suffix), the fail-closed durability ordering, and the `_source.messageId` precision check are all sound. The blockers are concentrated in the safety-field enrichment path and its verification gate.

---

## Blocking (P1)

### P1-1 — Fragile payload parse silently drops `visibles` → reader fail-OPEN, and the recon gate is blind to it

`internal/backfill/doc_enrich.go:54` (`extractVisibility`) decodes the payload into a strict struct:

```go
var v struct {
    SpaceID  string            `json:"space_id"`
    Visibles []json.RawMessage `json:"visibles"`
}
if err := json.Unmarshal(payload, &v); err != nil {
    return "", nil   // <-- whole payload abandoned
}
```

If `space_id` is present but **not a string** (e.g. a number or object in a dirty/legacy row), `json.Unmarshal` fails on the `space_id` field and aborts the *entire* decode — so a perfectly valid `visibles` array is dropped and the doc is written with empty `visibles`. Per the reader (`visibility.go:216`, empty `visibles` ⇒ no gate), that document becomes searchable by users who should not see it. I reproduced the abort behavior:

```
input={"space_id":123,"visibles":["u1","u2"]}  =>  spaceID="" visibles=[]
```

The existing test (`doc_enrich_test.go:99`) only covers dirty *elements inside* `visibles` (handled gracefully) — it does **not** cover a dirty `space_id` *type*, which is the path that aborts everything.

Aggravating factor: `internal/recon/sample.go:295` (`parsePayloadVisibility`) uses the **same** strict-struct parse. So the field-sampling reconciliation gate computes empty `visibles` on the MySQL side too, "matches" the empty ES doc, and reports OK. The safety net this PR adds cannot detect this exact fail-open. Since backfill is the *only* writer this period and historical enrichment is the whole point of the PR, a silent access-control fail-open that the bundled gate greenlights is blocking.

**Fix:** parse both fields leniently and independently (e.g. `map[string]json.RawMessage`, then decode each field, tolerating a malformed `space_id` without discarding `visibles`). Apply the same change to `recon/sample.go::parsePayloadVisibility` so the gate stays honest, and add a test with a non-string `space_id` co-occurring with a valid `visibles`.

### P1-2 — Live-ingestion gate is decoupled from actual field population (latent fail-open at the schema bump)

`internal/consumer/service.go:103` opens live ingestion solely on `LiveContractCarriesSafetyFields()`, which is just `searchmsg.SchemaVersion >= 2` (`doc.go:129`). But `DocFromMessage` (`doc.go:81`) never assigns `SpaceID`/`Visibles`/`MessageSeq`. The gate's "is it safe to write live?" signal (a version constant) is independent from "does the doc actually carry the safety fields" (the converter).

Today this is safe (version is pinned at 1, `Message` has no such fields). But the unblock condition is purely the version constant: when octo-lib bumps to `SchemaVersion=2` and the producer is enriched in phase 9, this gate opens **whether or not** `DocFromMessage` was updated to copy the new fields. If that wiring is forgotten, live ingestion silently writes empty `visibles` again — the precise fail-open this gate exists to prevent. The protection is real but its trigger is in the wrong place.

**Fix (cheap, do it now):** couple the gate to population, not just version. Either populate the fields in `DocFromMessage` behind the same constant, or add a guard/test that fails when `SchemaVersion >= SafetyFieldsSchemaVersion` but `DocFromMessage` does not emit the three fields — so the phase-9 bump cannot open the gate without wiring the converter.

---

## Non-blocking (P2)

- **P2-1 — Standalone `reconcile` sampling ignores DLQ exclusions.** `cmd/reconcile/main.go:135` calls `recon.CompareSamples` (no exclusion set), while the inline backfill path correctly uses `CompareSamplesExcluding` (`cmd/backfill/main.go:225`). On a window containing legitimate DLQ rows, the standalone gate will sample them, find no ES doc, and report `sample_missing > 0` — a false MISMATCH that can flake the `migrate-forward` alias cut (`scripts/forward-migrate.sh` step 3 depends on `make run-recon`). It fails *closed* (annoying, not dangerous), but it undermines the gate's credibility. Thread the known-DLQ id set through here too.

- **P2-2 — `Healthy()` can report OK with zero effective sample coverage.** `internal/recon/report.go:27` checks only `Mismatch == 0 && Missing == 0`. If every candidate row is excluded/skipped (`sample.go:124` returns a clean zero result), a *configured* field-sampling gate silently degrades to a no-op that still passes. Consider failing (or at least loudly warning) when sampling was requested (`sampleN > 0`) but `Sampled == 0`, so an empty sample can't masquerade as a clean one.

- **P2-3 — `payload.mergeForward.msgs` mapped as `object`, not `nested`** (`internal/esindex/mapping/octo-message.json:88`). Inner arrays get flattened, so you cannot correlate a specific child `type` with its `searchText` in the same sub-message. Fine for token recall; a limitation for structured queries. Acceptable if intentional — worth a one-line note in the mapping README.

- **P2-4 — `uint64` cast precedes the negative check** (`internal/backfill/doc_enrich.go:34`): `doc.MessageSeq = uint64(row.MessageSeq)` then `if row.MessageSeq < 0 { doc.MessageSeq = 0 }`. Functionally correct (the wrapped value is immediately overwritten), but reordering the guard first reads more safely.

---

## Coverage / blind spots

- **Verified locally:** reader visibles gate is fail-open on empty (`visibility.go:216`); p2p `spaceId` scope is fail-closed (`dsl.go:80`); octo-lib pin (`v0.0.0-20260614033907`) has `SchemaVersion=1` and `Message` lacks the three safety fields — so the live fail-closed gate is correctly armed this period; the dirty-`space_id` abort is reproduced empirically.
- **Could not verify from this repo:** the real type-cleanliness of the historical MySQL `payload.space_id` column (determines how often P1-1 actually fires); exact field alignment of the pushed gauge payload against `octo-server recon_metrics.go::ReconReport`; behavior under a live MySQL/OpenSearch cluster (no integration run).
- A second adversarial pass flagged two compile-break suspicions (`envInt` undefined in backfill; `dlqRecord.Reason` missing). Both are **false positives** from diff-only context — `envInt` is defined at `cmd/backfill/main.go:248` and `Reason` at `internal/backfill/dlq.go:21`; the package builds clean. Noting them so they aren't re-raised.

Requesting changes on P1-1 and P1-2; the P2 items can ride along or follow up.

Jerry-Xin (dismissed):
Correcting my earlier APPROVE on this head. After tracing the backfill visibility-parse path (which I had not fully traced before), I'm flipping to REQUEST_CHANGES on a real fail-OPEN. Credit to the reviewer who first surfaced the `space_id` parse concern.

**🔴 Blocking — `extractVisibility` drops the access-control list on a single malformed field (fail-OPEN on bad data)**

`internal/backfill/doc_enrich.go` `extractVisibility`:

```go
var v struct {
    SpaceID  string            `json:"space_id"`
    Visibles []json.RawMessage `json:"visibles"`
}
if err := json.Unmarshal(payload, &v); err != nil {
    return "", nil   // ← BOTH spaceID and visibles dropped
}
```

- `Visibles` is parsed leniently (per-element: non-string elements skipped) — good.
- But `SpaceID` is a strict `string`. If a payload's `space_id` is a non-string JSON type (number, object, upstream format drift), the **whole-struct `Unmarshal` fails** and the function returns `"", nil` — **`visibles` is silently emptied too**, even though the `visibles` data itself was valid.

Why that is a security failure: per `doc.go:21`, the reader treats **empty `visibles` as fail-OPEN** (the admin-only group-system-message whitelist gate is removed — ordinary members can search messages meant to be admin-only). So one row whose `space_id` is the wrong JSON type → empty `visibles` written → that row's access-control gate is silently dropped at the reader. A single-field type quirk collapses the visibility ACL. This is exactly the V3b fail-OPEN this PR's safety design exists to prevent, reintroduced via a parse asymmetry on the **active backfill path** (not behind the live-ingestion gate).

Severity: triggering depends on a non-string `space_id` actually appearing in payloads (dirty/drifted data), so it's not guaranteed-exploitable today — but it is a fail-OPEN-on-bad-data path in security-relevant code, while the safe posture everywhere else in this PR is fail-CLOSED. Empty `visibles` must never be the silent result of a parse error.

Fix (fail-closed): if visibility can't be parsed, the row should go to DLQ/permanent (skip the write), not be written with empty `visibles`. At minimum, parse `space_id` as `json.RawMessage` (or a tolerant type) so its JSON type can't nuke `visibles`, AND treat a genuine visibility-parse failure as fail-closed (skip/DLQ), not "no constraints." Add a test: a payload with a non-string `space_id` (valid `visibles`) must NOT yield empty `visibles` — it should preserve `visibles` or fail-closed.

**💬 Non-blocking (carried over / still valid)**
- 🟡 The live-ingestion safety gate keys live-write permission on `searchmsg.SchemaVersion >= SafetyFieldsSchemaVersion`, but `DocFromMessage` doesn't copy `spaceId/visibles/messageSeq`. Safe at v1 (gate blocks writes; contract lacks the fields; backfill enriches from MySQL). But the gate's version check is an unenforced proxy for "DocFromMessage maps the safety fields": when the contract bumps to v2, if `DocFromMessage` isn't updated in lockstep the gate opens and live docs ship without `visibles` → fail-OPEN returns. Add a CI test that's red until `DocFromMessage` copies the three fields for a v2-shaped message, so the v2 bump can't merge without wiring the mapping.
- 🟡 `forward-migrate.sh`: the `step3_switch_alias` `_aliases` remove-then-add will error if the old alias is absent — make the remove idempotent (`ignore_unavailable` / tolerate-missing). Also the mapping file itself carries the read alias; a raw `PUT` of it would create an index that's live-aliased on creation (mixed read). The script strips it via `jq 'del(.aliases)'`, so the script path is safe — but consider removing the alias from the mapping file itself so a direct `PUT` is safe by default.
- 🟡 backfill inline `-reconcile` still runs only the count gate, not the new field-level sample gate.

**✅ Highlights (unchanged)**
- Lint blocker from the prior round is fixed; CI Build/Vet/Lint/Test green.
- OS doc contract converges to reader shape byte-for-byte (camelCase `messageId` long, nested `payload.text.content`).
- snowflake precision safe: reconcile reads typed `_source.messageId` int64, not the float64 Sort.
- Non-numeric `message_id` is a permanent failure (not coerced to 0); `Bulk()` correctly excludes permanent items from the batch-failure retry set.

Net: the Lint blocker is cleared, but the `space_id`-parse fail-OPEN on the backfill path is a real should-fix-before-merge on the access-control field. Holding REQUEST_CHANGES on that. Remaining merge gate also includes `check-sprint` (PR needs `Closes #<issue>`).

Jerry-Xin (dismissed):
The PR is relevant to `octo-search-indexer`, but the standalone reconciliation gate has a blocking correctness issue.

**🔴 Blocking**
- 🔴 Critical: Standalone `cmd/reconcile` subtracts the DLQ count for count reconciliation, but its field-level sample path has no DLQ exclusion set. In [cmd/reconcile/main.go](/Users/mlamp/.openclaw/tmp/pr-review-oa4lrgqu/repo/cmd/reconcile/main.go:131), it calls `recon.CompareSamples(...)`, which samples MySQL rows and treats any missing ES document as `sample_missing` in [internal/recon/sample.go](/Users/mlamp/.openclaw/tmp/pr-review-oa4lrgqu/repo/internal/recon/sample.go:137). That is correct for ordinary missing docs, but false-fails legitimate DLQ rows: malformed payloads, non-numeric IDs, or permanent ES rejects are intentionally absent from ES and already accounted for by `-dlq`. The inline backfill reconcile handles this via `CompareSamplesExcluding`, but the documented mandatory standalone gate does not. As written, `make run-recon RECON_DLQ=<count>` can exit 2 even when the backfill is correct, whenever a DLQ row lands in the deterministic sample. This contradicts the mandatory standalone workflow in [README.md](/Users/mlamp/.openclaw/tmp/pr-review-oa4lrgqu/repo/README.md:106) and the alias-switch precondition in [docs/forward-migration-v1.9.md](/Users/mlamp/.openclaw/tmp/pr-review-oa4lrgqu/repo/docs/forward-migration-v1.9.md:64). Fix by giving standalone reconcile access to DLQ message IDs, for example via a spill-file/path flag, and use `CompareSamplesExcluding` there too.

**✅ Highlights**
- The writer-side document shape is coherently moved to the reader contract, with full-precision numeric `messageId`.
- Backfill enrichment correctly treats broken visibility parsing as fail-closed.
- The live-ingestion safety gate is a good guard against writing fail-open docs before the Kafka contract carries the required fields.
- `go test ./...` passes locally.

Jerry-Xin (approved):
Summary: The PR is in-scope for `octo-search-indexer` and the core writer/backfill/reconcile changes are consistent with the reader-aligned document contract.

💬 Non-blocking

🟡 Warning — `scripts/forward-migrate.sh:57` uses `Long.parseLong(ctx._source.remove('message_id'))`, so a single legacy non-numeric `message_id` aborts the whole reindex step. The Go writer/backfill paths now route these to DLQ, but the reindex-only migration path has no equivalent skip/DLQ accounting. If production legacy data is guaranteed numeric this is acceptable; otherwise add a precheck, filtered reindex, or documented hard-stop handling.

🟡 Warning — `cmd/reconcile/main.go:110-150` does not refresh the target index before count/sample checks, while inline backfill reconcile does at `cmd/backfill/main.go:184-188`. Standalone `make run-recon` may false-fail immediately after backfill or when `refresh_interval=-1`. Consider adding an optional/default refresh before standalone reconciliation.

🔵 Suggestion — `harness/verify/main.go:90` sends a string term against the `messageId` long field. OpenSearch often coerces this, but using a numeric JSON value would make the harness stricter about the new contract.

✅ Highlights

- `internal/esindex/writer.go` cleanly separates `Bulk` conversion from `BulkDocs`, preserving per-item permanent errors for non-numeric IDs.
- `internal/backfill/doc_enrich.go` handles visibility parsing defensively and avoids silently dropping ACLs.
- `internal/recon/sample.go` adds useful field-level drift detection, including `_source.messageId` precision.
- The live ingestion safety gate plus v2 reflection guard are good protection against future fail-open regressions.

Verification: `go test ./...` passes.

lml2468:
## ✅ Re-review @5746484 — standalone-reconcile DLQ false-fail FIXED. Approving.

The migration-gate false-fail I held my vote on (Steve's catch, I reproduced the logic) is now fixed correctly. Verified at 5746484:

### 🔴→✅ standalone reconcile now excludes DLQ rows from the sample gate
- New -dlq-spill-dir flag / RECON_DLQ_SPILL_DIR env (main.go:59): standalone reconcile reads the backfill DLQ spill file.
- backfill.LoadDLQMessageIDsInWindow(dlqDir, from, to) (main.go:146) reconstructs the same exclusion set the inline path builds.
- standalone now calls CompareSamplesExcluding(..., excluded) — consistent with the inline backfill gate (cmd/backfill/main.go:225).
- So legit DLQ rows (malformed payload / non-numeric id / this PR's fail-closed→DLQ rows) are no longer counted as Sample.Missing → no false os.Exit(2) on a correct backfill. The mandatory migration gate won't false-fail.

### Both blockers from this review now closed
- extractVisibility fail-OPEN → fixed (top-level map[string]json.RawMessage, space_id type-tolerant doesn't collapse visibles, corrupt visibles → DLQ fail-closed; verified, reproduction closed).
- standalone reconcile DLQ false-fail → fixed (this; exclusion set wired, parity with inline).
- Lint green, build + go test ./internal/recon ./internal/backfill green.

### Remaining (non-blocking 🟡)
- DocFromMessage v2 latent coupling (SchemaVersion as proxy for safety fields being mapped) — fix-before-v2; recommend a CI test red until the mapping is wired. Not a v1 blocker.
- migration painless on non-numeric legacy _id; standalone reconcile doesn't force-refresh ES before sampling — minor follow-ups.

### Verdict
Both real security/correctness blockers (access-control fail-OPEN + migration-gate false-fail) genuinely fixed and verified; doc-contract convergence, precision, migration safety, recon sampling sound. LGTM. Approving.

lml2468 (approved):
Re-approve @5746484: standalone-reconcile DLQ false-fail fixed (new -dlq-spill-dir + LoadDLQMessageIDsInWindow rebuilds exclusion set, standalone now uses CompareSamplesExcluding, parity with inline gate, so legit DLQ rows no longer false-fail as sample_missing). With the verified extractVisibility fail-OPEN fix, both blockers closed. build + go test green, Lint green. Top follow-up: CI test red until DocFromMessage maps v2 safety fields. LGTM.

PR #17, opened 2026-06-20
feat: add standalone searchetl-producer service (MySQL→Kafka polling ETL)

yujiawei:
## What

Adds a standalone polling ETL **producer** service to this repo, packaged as a new binary alongside `es-indexer`. It reads the MySQL `message` shard tables on a slow-cursor tick, enriches each row into the octo-lib `searchmsg.Message` contract (fail-closed visibility), and produces it to Kafka — where the existing `es-indexer` consumer picks it up and writes OpenSearch.

```
message shards (MySQL) → [searchetl-producer: poll → enrich → Kafka] octo.message.v1(.dlq)
  → es-indexer consumer → OpenSearch
```

## Why this shape (slim port, not a module lift)

The producer logic exists today inside the server's `searchetl` module, but that module carries heavy dependencies (dbr, opentracing globals) unsuitable for this slim, single-purpose image. This PR ports the logic as **pure functions** in the target repo's existing style, with no new heavy deps:

- **Payload extraction** aligns with this repo's `internal/backfill` extract; the **shared octo-lib `searchmsg.ExtractVisibility`** is the single source of truth for fail-closed visibility (the same vectors lock both producer and backfill).
- **Redis distributed lock** is self-contained (`go-redis` + octo-lib `BuildTLSConfig`, SET NX EX + Lua CAS-DEL/CAS-PEXPIRE) — no dependency on the server's redis package.
- **Contract**: octo-lib `searchmsg.Message` (SchemaVersion 2), already a dependency here.

## Boundary vs es-indexer

`es-indexer` is the long-running Kafka→OpenSearch **consumer**; this is the MySQL→Kafka **producer**. They share only the octo-lib contract and the Kafka topics. `backfill` is the one-shot historical loader (bypasses Kafka); this producer is the realtime/incremental path.

## Zero production risk by construction

The producer is **opt-in**: with `PRODUCER_ENABLED` unset/false (or brokers/DSN/Redis unset) it idles to signal and connects to nothing. No existing scheduler, deployment, or traffic is touched.

## Scope

- `cmd/searchetl-producer` + `internal/producer`: slow-cursor scheduler, MySQL keyset source + cursor table, enrichment (visibility fail-closed), Kafka writer, Redis lock.
- Config maps env semantics (tick/lag/batch, source DB, Kafka brokers, Redis), with **explicit connection-pool config** (no bare driver defaults).
[...]

PR #18, opened 2026-06-20
feat(producer): forensic DLQ envelope + run first tick on start

yujiawei:
## What

Two non-blocking refinements to the standalone `searchetl-producer` raised in
the #17 review, plus a small idle-posture fix so the service can be deployed
disabled-by-default. **No change to the body-topic contract or the correctness
path** (stability gate, cursor CAS, produce-before-advance, Redis run-lock,
idempotent sink).

### 1. Producer-specific forensic DLQ envelope 🟡
Dead-lettered rows were serialized as a bare `searchmsg.Message` (the body
contract), which dropped the context needed to triage/replay a poison row and
did not match the `{reason, source locator, raw bytes, timestamps}` vocabulary
the consumer (`internal/consumer`) and backfill (`internal/backfill`) DLQ
records already use.

- New `DLQEnvelope` carrying `reason`, `shard_table`, `source_id`, `message_id`,
  `raw_payload`, and timestamps.
- `extractMessage` now also returns a machine-readable reason
  (`producer_payload_unparseable` vs `producer_visibility_untrusted`) so triage
  can tell a malformed payload apart from a fail-closed ACL drop.
- The DLQ stream is terminal (es-indexer never reads it), so the envelope is free
  to diverge from the indexable body shape. `planChunk` stays a pure function:
  `produced_at` is stamped by the Kafka sink at produce time.
- DLQ write failures still fail the whole chunk (never silently drop); the Kafka
  key stays `message_id`.

### 2. Scheduler runs its first tick on start 🔵
The scheduler waited a full ticker interval (default 60s) before its first round
— slow first-light on deploy and a long blind window at the runtime cut-over
(the cursor is expected to advance within seconds). It now runs one round
immediately on `Start`, then ticks on cadence. The run-lock + cursor CAS +
idempotent sink make the early round safe; it respects ctx cancellation so a
fast shutdown right after `Start` fires nothing.

### 3. Idle-posture health endpoints (enables opt-in deploy)
For the "deployed but off" posture (`PRODUCER_ENABLED` unset/false), the binary
now serves `/healthz` + `/readyz` (both 200, idling-is-ready) so a disabled pod
does not crashloop under HTTP liveness/readiness probes. `LoadConfig` returns
the master switch alone; `main` fail-fasts when `PRODUCER_ENABLED=true` but the
config is incomplete (a requested producer must not silently idle "ready" and
no-op a cut-over). Still connects to no backend while idle.

## Tests
- DLQ reason mapping + envelope shape/marshal; `planChunk` DLQ-envelope assertions.
- Scheduler immediate-first-tick; idle obs health/ready.
- `LoadConfig` master-switch-only semantics + requested-but-misconfigured fails Validate.
- e2e producer harness now decodes + asserts the forensic envelope on the DLQ topic.

## Review
- `go build` / `go vet` / `go test -race -shuffle=on ./...`: all green
- golangci-lint v2.2.0: 0 issues
- Independent model review (gpt-5.5, high): 2 findings, both fixed and confirmed RESOLVED on re-review.

## Pairs with
The octo-deployment change that wires this service into the `search` profile
(opt-in, OFF by default). **Merge this first** so the published image carries the
forensic DLQ envelope + first-tick behavior the deployment references.

### DIFF
--- a/cmd/searchetl-producer/main.go
+++ b/cmd/searchetl-producer/main.go
@@ -24,6 +24,7 @@ package main
 
 import (
 	"context"
+	"fmt"
 	"log"
 	"os"
 	"os/signal"
@@ -48,14 +49,24 @@ func main() {
 }
 
 func run(ctx context.Context) error {
-	cfg, enabled := producer.LoadConfig()
-	if !enabled {
-		log.Printf("searchetl-producer: PRODUCER_ENABLED not true (or brokers/DSN/Redis unset); idling (no backend connection)")
+	cfg, requested := producer.LoadConfig()
+	if !requested {
+		log.Printf("searchetl-producer: PRODUCER_ENABLED not true; idling (no backend connection)")
+		// Serve health endpoints even while idling so an orchestrator's
+		// liveness/readiness probes do not crashloop a deliberately-disabled
+		// pod (the opt-in "deployed but off" posture). Both /healthz and /readyz
+		// report 200: per the plan, "idling normally" is a ready state — it lets
+		// the opt-in deploy roll out cleanly without receiving any traffic (the
+		// producer exposes no Service). It connects to no backend.
+		serveIdleObs(ctx, cfg.ObsAddr)
 		<-ctx.Done()
 		return nil
 	}
+	// 🔴 PRODUCER_ENABLED=true means the operator intends this to run. Missing
+	// DSN/brokers/Redis/lag is then a hard error (fail-fast crashloop), NOT a
+	// silent idle — otherwise a cut-over would no-op while reporting healthy.
 	if err := cfg.Validate(); err != nil {
-		return err
+		return fmt.Errorf("searchetl-producer: PRODUCER_ENABLED=true but config invalid (refusing to idle on a requested producer): %w", err)
 	}
 
 	// Source DB with an explicit connection pool (no bare driver defaults).
@@ -138,4 +149,30 @@ func run(ctx context.Context) error {
 	}
 	sched.Stop()
 	return nil
+}
+
+// serveIdleObs starts the observability HTTP server for the idle (disabled)
+// posture: /healthz and /readyz both return 200. Per the plan, "idling
+// normally" is a ready state — this keeps a deliberately-disabled pod
+// (PRODUCER_ENABLED=false, the opt-in "deployed but off" default) from
+// crashlooping under an orchestrator's HTTP probes and lets the opt-in deploy
+// roll out cleanly. It connects to no backend (ready check is nil; the producer
+// exposes no Service so a ready idle pod receives no traffic). A graceful
+// shutdown stops it on ctx cancel.
+func serveIdleObs(ctx context.Context, addr string) {
+	if addr == "" {
+		return
+	}
+	obs := producer.NewObsServer(addr, nil, nil)
+	obs.SetReady(true) // idling normally is a ready state (no backend dialed)
+	obs.Start(log.Printf)
+	go func() {
+		<-ctx.Done()
+		sctx, cancel := context.WithTimeout(context.Background(), 5*time.Second)
+		defer cancel()
+		if serr := obs.Shutdown(sctx); serr != nil {
+			log.Printf("searchetl-producer: idle obs shutdown: %v", serr)
+		}
+	}()
+	log.Printf("searchetl-producer: idle observability server on %s (/healthz + /readyz 200, idle)", addr)
 }
--- a/cmd/searchetl-producer/main_test.go
+++ b/cmd/searchetl-producer/main_test.go
@@ -1,0 +1,50 @@
+package main
+
+import (
+	"context"
+	"fmt"
+	"net"
+	"net/http"
+	"testing"
+	"time"
+)
+
+// TestServeIdleObs_HealthAndReady verifies the idle (disabled) posture serves
+// both /healthz and /readyz as 200, so a deliberately-off pod does not crashloop
+// under HTTP liveness/readiness probes.
+func TestServeIdleObs_HealthAndReady(t *testing.T) {
+	// Grab a free localhost port.
+	ln, err := net.Listen("tcp", "127.0.0.1:0")
+	if err != nil {
+		t.Fatalf("listen: %v", err)
+	}
+	addr := ln.Addr().String()
+	if cerr := ln.Close(); cerr != nil {
+		t.Fatalf("close probe listener: %v", cerr)
+	}
+
+	ctx, cancel := context.WithCancel(context.Background())
+	defer cancel()
+	serveIdleObs(ctx, addr)
+
+	base := fmt.Sprintf("http://%s", addr)
+	deadline := time.Now().Add(3 * time.Second)
+	for _, path := range []string{"/healthz", "/readyz"} {
+		var got int
+		for time.Now().Before(deadline) {
+			resp, gerr := http.Get(base + path) //nolint:noctx // short-lived test probe
+			if gerr != nil {
+				time.Sleep(20 * time.Millisecond)
+				continue
+			}
+			got = resp.StatusCode
+			if cerr := resp.Body.Close(); cerr != nil {
+				t.Fatalf("close body: %v", cerr)
+			}
+			break
+		}
+		if got != http.StatusOK {
+			t.Fatalf("idle %s = %d, want 200", path, got)
+		}
+	}
+}
--- a/harness/producer/README.md
+++ b/harness/producer/README.md
@@ -16,9 +16,11 @@ seeds a controlled message fixture (`seed.sql`), and runs the Go verifier
 1. **poll → enrich → Kafka** — stable rows land on the body topic with the right
    enrichment (text body, `visibles` whitelist, `raw_excluded` for Signal/media),
    and genuine anomalies + fail-closed visibility violations land on the DLQ
-   topic. This is a **field-level** assertion (not a bare doc count): it checks
-   the actual contract contents, including the #1124 fail-closed guard that the
-   empty-`visibles` row must route to DLQ, never to the body topic.
+   topic as forensic **DLQ envelopes** (reason + raw payload + source locator,
+   not bare body contracts). This is a **field-level** assertion (not a bare doc
+   count): it checks the actual contract contents, including the #1124
+   fail-closed guard that the empty-`visibles` row must route to DLQ, never to
+   the body topic.
 2. **Redis lock mutual exclusion** — two concurrent `RunIncremental` calls
    sharing the same lock key never both produce; exactly one wins, the other
    skips.
--- a/harness/producer/main.go
+++ b/harness/producer/main.go
@@ -191,15 +191,15 @@ func run() error {
 	if err != nil {
 		return fmt.Errorf("drain main topic: %w", err)
 	}
-	dlqMsgs, err := drainTopic(ctx, cfg.Brokers, cfg.DLQTopic, 2)
+	dlqEnvelopes, err := drainDLQ(ctx, cfg.Brokers, cfg.DLQTopic, 2)
 	if err != nil {
 		return fmt.Errorf("drain dlq topic: %w", err)
 	}
 	if len(mainMsgs) != 4 {
 		return fmt.Errorf("main topic: want 4 messages, got %d", len(mainMsgs))
 	}
-	if len(dlqMsgs) != 2 {
-		return fmt.Errorf("dlq topic: want 2 messages, got %d", len(dlqMsgs))
+	if len(dlqEnvelopes) != 2 {
+		return fmt.Errorf("dlq topic: want 2 messages, got %d", len(dlqEnvelopes))
 	}
 
 	// Assert enrichment correctness on the main stream.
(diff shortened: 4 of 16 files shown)


lml2468 (approved):
## ✅ Approve:#17 两条 DLQ/scheduler 🟡 的正解实现,核验通过

`feat(producer): forensic DLQ envelope + run first tick on start`(16 文件 +526-85)。落地 #17 的两条 follow-up。基线 build+test+`-race` 全过。

### ✅ dlq.go forensic envelope
新 `DLQEnvelope`(reason/shard_table/source_id/message_id/raw_payload/timestamps)替代裸 searchmsg.Message 进 DLQ,留下 triage/replay 上下文。词汇对齐 consumer/backfill DLQ schema(跨三 DLQ 同一 triage 工具),SchemaVersion 独立。`newDLQEnvelope` 纯函数(ProducedAt 后盖)。

### ✅ DLQ reason 接线精确(实测)
- `extract.go:112` payload 解析失败 → `dlqReasonPayloadUnparseable`
- `extract.go:120` visibility 不可信 → `dlqReasonVisibilityUntrusted`(#1124 fail-closed 保留)
精确对应,非都标一个。`chunk.go:47` 封装。**fail-closed visibility 逻辑未破坏**,只是带了 envelope。

### ✅ scheduler 首 tick — 正解
`loop` 进循环前先 `if ctx.Err()==nil { s.tick(ctx) }` 立即跑一轮,避免 boot 空等 60s / cut-over 秒级接管;安全靠 run-lock + CAS + 幂等 + 尊重 ctx 取消。正是 #17"首次延迟" 🟡 正解。

### 🔵 nit(非阻塞)— DLQ envelope 空 MessageID 作 key
`kafka.go:97` DLQ 用 `Key:[]byte(MessageID)`,但 envelope MessageID "may be empty for malformed row"。malformed row(正是进 DLQ 的)空 key 概率更高 → 挤同一分区。但不影响正确性(DLQ 终端流、流量极低、triage 按 source_id 定位)。与 yujiawei #17 空 key nit 同源,follow-up 一并收(空则用 source_id 作 key)。

### 结论
envelope 设计对 + reason 接线精确 + fail-closed 未破 + scheduler 首 tick 正解 + build/test/race 全过 —— **approve**。唯一 🔵 DLQ 空 key 非正确性,follow-up。

> 作者 @yujiawei(非我账号),核验无阻塞,投 approve。

— Octo-PR 🐙(clone+build+`-race`+读 dlq/extract/chunk/scheduler/kafka;实测 reason 双分支 + scheduler 首 tick + fail-closed 未破)

mochashanyao (approved):
[Octo-Q · automated review]

---

# octo-search-indexer PR#18 Review Report
**PR:** https://github.com/Mininglamp-OSS/octo-search-indexer/pull/18
**Head SHA:** `571463a3b78036b9cb2935ed15ec32f986f90ea2`
**Title:** feat(producer): forensic DLQ envelope + run first tick on start

---

## 1. Verification Summary

| Item | Status | Evidence |
|------|--------|----------|
| DLQ envelope data flow (build → plan → produce) | ✅ | `dlq.go:78-91` → `chunk.go:44-48` → `kafka.go:82-103` |
| extractMessage 3-value return (msg, outcome, reason) | ✅ | `extract.go:84` signature; all 6 return sites carry correct reason |
| planChunk purity (no clock) | ✅ | `dlq.go:88` ProducedAt left 0; `kafka.go:87` stamps at produce time |
| Sink interface migration (Message → DLQEnvelope) | ✅ | `kafka.go:20` interface; `etl_test.go:78,94,267` fakes; `harness/producer/main.go:280` countingSink |
| Scheduler immediate first tick | ✅ | `scheduler.go:72-74` ctx.Err() guard; `scheduler_test.go:62-103` two tests |
| Idle obs server (healthz/readyz) | ✅ | `main.go:163-178`; `main_test.go:15-50` |
| LoadConfig master-switch-only semantics | ✅ | `config.go:169` returns requested=enabled alone; `config_test.go:8-44` |
| PRODUCER_ENABLED=true + misconfig → fail-fast | ✅ | `main.go:68-70` wraps Validate error with explicit context |
| Harness e2e DLQ envelope assertions | ✅ | `harness/producer/main.go:231-268` decodes envelopes, asserts forensic fields |

## 2. Findings

### No P0 or P1 findings.

All three changes (DLQ envelope, first tick, idle obs) are well-implemented with thorough tests and correct data flow.

### P2 Observations (non-blocking)

**P2-1: ProduceDLQ mutates input slice elements**
- `kafka.go:88`: `envelopes[i].ProducedAt = now` writes back into the caller's slice.
- Currently safe: `plan.dlq` is consumed once in `runChunk` and the `chunkPlan` is discarded after return.
- Risk: if a future caller reuses the slice (e.g., for retry/replay), the stamped `ProducedAt` would be stale.
- **Suggestion:** Consider copying the envelope before mutation, or document the mutation contract on the `Sink` interface.

**P2-2: Empty Kafka key for DLQ envelopes with empty MessageID**
- `kafka.go:94`: `Key: []byte(envelopes[i].MessageID)` — if MessageID is empty (the `dlq.go:37` comment acknowledges this is possible for malformed rows), the key is `[]byte("")`.
- With `kafka.Hash{}` balancer, an empty key hashes to a consistent partition, so this is not a correctness issue.
- However, the PR description states "Kafka key stays message_id (so DLQ records for the same row land in the same partition)" — for empty-MessageID rows, co-location by SourceID+ShardTable would be more useful for replay tooling.
- **Suggestion:** Consider falling back to `fmt.Sprintf("%s:%d", ShardTable, SourceID)` as key when MessageID is empty.

**P2-3: Idle obs server ListenAndServe error is fire-and-forget**
- `main.go:167`: `obs.Start(log.Printf)` launches the HTTP server in a goroutine; if `ListenAndServe` fails (e.g., port already bound), the error is logged asynchronously but `serveIdleObs` has already returned.
- The process continues to idle (`<-ctx.Done()`), which is correct behavior — but the operator may miss the log line and wonder why probes fail.
- **Suggestion:** Minor — consider a brief sleep + log check, or document this as expected behavior.

## 3. Suggestions

- P2-1 and P2-2 above are the main improvement opportunities.
- The code is heavily commented (good for this domain), but some comments are quite long (e.g., `main.go:52-60` block comment). Consider whether the comment density aids or hinders future maintenance.

## 4. Additional Findings

None beyond the P2 observations above.

## 5. Data Flow Tracing

### DLQ Envelope: source row → Kafka

```
srcMessageRow (DB row)
  → extractMessage(row) [extract.go:84]
    → returns (msg, outcomeDLQ, reason)
      reason = dlqReasonPayloadUnparseable (bad JSON/empty map) [extract.go:108]
      reason = dlqReasonVisibilityUntrusted (ACL fail-closed) [extract.go:120]
  → planChunk(table, rows, cutoff) [chunk.go:34]
    → newDLQEnvelope(table, row, reason) [chunk.go:46]
      → DLQEnvelope{SchemaVersion=1, Reason, ShardTable=table, SourceID=row.ID,
         MessageID=row.MessageID, ChannelID, ChannelType, RawPayload=row.Payload,
         CreatedAt=row.CreatedUnix, MsgTimestamp=row.Timestamp, ProducedAt=0}
    → appended to plan.dlq [chunk.go:46]
  → runChunk [etl.go:170]
    → sink.ProduceDLQ(ctx, plan.dlq) [etl.go:186]
      → KafkaProducer.ProduceDLQ [kafka.go:82]
        → stamps ProducedAt = time.Now().Unix() [kafka.go:88]
        → json.Marshal(envelopes[i]) [kafka.go:89]
        → kafka.Message{Key: []byte(MessageID), Value: marshaled} [kafka.go:94]
        → dlqWriter.WriteMessages(ctx, kmsgs...) [kafka.go:97]
```

All data flows verified: every field consumed at the Kafka sink is correctly populated from the source row. ProducedAt is correctly deferred to the sink (planChunk stays pure). ✓

### Scheduler first tick: Start → loop → tick

```
Scheduler.Start() [scheduler.go:52]
  → go loop(ctx) [scheduler.go:57]
    → time.NewTicker(interval) [scheduler.go:64]
    → if ctx.Err() == nil → s.tick(ctx) [scheduler.go:72-74]
      → tickFn(ctx, tables) [scheduler.go:87]
        → ETL.RunIncremental [etl.go:82]
          → runLocked (Redis lock) → runTick → runChunk per table
    → for { select { ctx.Done() / t.C } } [scheduler.go:75-82]
```

Verified: immediate tick respects ctx cancellation; run-lock prevents concurrent execution with subsequent ticker ticks; Stop() cancels ctx and joins with 30s timeout. ✓

### LoadConfig: enabled → requested separation

```
LoadConfig() [config.go:140]
  → cfg assembled from env vars
  → requested = strings.EqualFold(os.Getenv("PRODUCER_ENABLED"), "true") [config.go:169]
  → return cfg, requested

main.run() [main.go:51]
  → if !requested → serveIdleObs(ctx, cfg.ObsAddr) → <-ctx.Done() [main.go:53-61]
  → if requested → cfg.Validate() → fail-fast on error [main.go:68-70]
    → continue with DB/Redis/Kafka setup
```

Verified: the old behavior (enabled = flag AND config complete) is correctly split into two stages. A misconfigured-but-requested producer now crashloops instead of silently idling. ✓

- **C1 — Dual-path parity:** N/A. No symmetric add/remove or create/delete paths in this PR. The Sink interface change (ProduceDLQ signature) is consistently applied across all implementations (KafkaProducer, fakeSink, blockingSinkT, countingSink). ✓
- **C2 — Control-flow ordering / nested reuse:** Clear. The immediate tick runs once before the ticker loop; run-lock prevents concurrent execution. ProduceDLQ stamps ProducedAt on all envelopes before any Kafka write; marshal failure aborts the entire batch before WriteMessages. No double-action risk. ✓
- **C3 — Authorization boundary:** N/A. No auth/permission/jail changes. This is an internal ETL pipeline.
- **C4 — Authorization lifecycle / container-member cascade:** N/A. No auth changes.
- **C5 — Build ≠ runtime:** N/A. No build artifacts, extensions, CLI packaging, or relative URL changes.
- **C6 — Governance/policy document consistency:** N/A. No governance or policy documents changed.

## 7. Cross-round Blocker Recheck (R6)

---

**[Octo-Q] verdict: APPROVE**

All three changes are well-implemented with correct data flow, thorough tests, and clear design rationale. No P0/P1 findings. Three P2 observations (input slice mutation, empty Kafka key edge case, fire-and-forget obs error) are non-blocking improvements.

Jerry-Xin (approved):
The PR is in-scope for `octo-search-indexer` and the implementation is ready to merge; I found no blocking correctness, security, or contract issues.

**💬 Non-blocking**
- 🔵 Suggestion: The fast-stop guarantee is slightly overstated. `internal/producer/scheduler.go:72` checks `ctx.Err()` before the immediate tick, but `Stop()` can still race after that check and allow one tick; `internal/producer/scheduler_test.go:97` already acknowledges this. Consider aligning the comment/PR wording with the actual “at most one immediate tick may race” behavior.
- 🟡 Warning: Idle observability startup failures are only logged asynchronously. In `cmd/searchetl-producer/main.go:166` and `internal/producer/obs.go:71`, a bind failure leaves the disabled process alive but not serving probes. Since idle readiness is the point of this path, consider making `Start` return startup errors or adding a bind-failure test.
- 🔵 Suggestion: The top comment in `cmd/searchetl-producer/main.go:20` still says missing brokers/DSN/Redis causes idle behavior. After this PR, `PRODUCER_ENABLED=true` with incomplete config fail-fasts at `cmd/searchetl-producer/main.go:68`, so the comment should be updated.

**✅ Highlights**
- DLQ records now carry useful forensic context without changing the body-topic contract.
- `planChunk` remains pure, with `ProducedAt` stamped at Kafka write time.
- The enabled-vs-requested config split correctly prevents a requested producer from silently idling.
- Test coverage was added for envelope shape, reason mapping, idle health, scheduler first tick, and config semantics.

Verified locally: `go test ./...` and `go test -race ./internal/producer ./cmd/searchetl-producer` both pass.

yujiawei (commented):
# Code Review — PR #18 (octo-search-indexer)

**Head reviewed:** `571463a3b78036b9cb2935ed15ec32f986f90ea2`
**Base / merge-base:** `ec49d70` (single commit on top of `main`)
**Build/vet/test:** `go build ./...`, `go vet ./...`, `go test ./internal/producer/... ./cmd/...` — all green locally (go1.26.3).

## Verdict: APPROVED

No P0/P1 (correctness / security / data-loss / build-break) findings. Three independent perspectives (line-level + cross-file long-context + my own re-read of the diff) surfaced only P2/nit-level, non-blocking items. The correctness path that carries the production risk — stability gate, cursor CAS, produce-before-advance ordering, Redis run-lock, idempotent ES sink — is untouched and was re-verified against this head.

---

## 1. What the PR does (and what it does NOT touch)

Three scoped refinements to the standalone `searchetl-producer`:

1. **Forensic DLQ envelope** — dead-lettered rows now serialize as a `DLQEnvelope` (`reason`, `shard_table`, `source_id`, `message_id`, `raw_payload`, timestamps) instead of a bare `searchmsg.Message`. `extractMessage` now also returns a machine-readable reason (`producer_payload_unparseable` vs `producer_visibility_untrusted`).
2. **Scheduler first-tick on start** — one round runs immediately on `Start`, then on cadence, instead of idling a full ticker interval.
3. **Idle-posture health endpoints** — a disabled pod (`PRODUCER_ENABLED` unset/false) serves `/healthz` + `/readyz` (200) so it does not crashloop under HTTP probes; `LoadConfig` now returns the master switch alone and `main` fail-fasts when enabled-but-misconfigured.

The body-topic contract, ACL fail-closed semantics, and the cursor-advance ordering are unchanged.

## 2. Verification results

- ✅ **DLQ envelope is sound.** `planChunk` stays a pure function — `ProducedAt` is correctly deferred to the Kafka sink (`kafka.go:88-91`), not stamped at extraction; `newDLQEnvelope` (`dlq.go:78`) takes no clock. The reason mapping is correct: parse-failure/empty-map → `producer_payload_unparseable`, untrusted visibility → `producer_visibility_untrusted` (`extract.go:108,120`), and non-DLQ paths return `""` (`dlq_test.go` asserts this). DLQ write failure still fails the whole chunk and leaves the cursor un-advanced (`etl.go:189-191`) — no silent drop. ✅
- ✅ **DLQ stream is genuinely terminal.** Confirmed the es-indexer consumer reads the body topic and only *writes* its own DLQ sink — it never consumes the producer DLQ topic. So diverging the DLQ shape from the indexable body contract is safe, as the PR description claims.
- ✅ **Schema vocabulary is consistent** with `internal/consumer` `dlqRecord` and `internal/backfill` `dlqRecord` (`{reason, source locator, raw bytes, timestamps}`), with a producer-namespaced `reason` so a shared triage tool can distinguish the three sources.
- ✅ **First-tick is safe.** The `ctx.Err() == nil` guard before the immediate `s.tick` (`scheduler.go:69`) correctly suppresses the early round on a fast `Start`+`Stop`. Safety of the extra early round rests on the run-lock + CAS + idempotent sink, all verified present (`etl.go:77-86`).
- ✅ **Idle-posture nil callbacks are safe.** `serveIdleObs` passes `nil` metrics and `nil` ready-check into `NewObsServer`; both are guarded (`obs.go:45` `if ready != nil`, `obs.go:58` `if metrics != nil`). A disabled pod serves `/healthz` + `/readyz` 200 and dials no backend.
- ✅ **Master-switch separation is correct.** `LoadConfig` returns `requested` = `PRODUCER_ENABLED==true` alone; `main` then `Validate()`s and fail-fasts on a requested-but-misconfigured producer (`main.go:69-72`), so a cut-over can never silently idle "ready" while producing nothing. Good defensive call.
- ✅ **Tests + e2e harness updated** to decode and assert the forensic envelope (schema_version, reason, raw payload, source locator, produced_at) on the DLQ topic.

## 3. Findings (all non-blocking)

### P2 — Bound / be explicit about `raw_payload` size on the DLQ topic
*(Independently raised by both review legs — highest-confidence item.)*

`newDLQEnvelope` copies `row.Payload` verbatim into the envelope (`dlq.go:87`), which `encoding/json` then base64-encodes (≈ +33%). For a genuinely large poison row, the resulting DLQ record can exceed the Kafka writer's `BatchBytes` / broker `message.max.bytes` (both default 1 MiB). `ProduceDLQ` would then return an error and `runChunk` intentionally leaves the cursor un-advanced — so the producer retries the same chunk forever and **head-of-line-blocks that shard** on a single poison row.

This is an edge case (it needs a very large malformed payload), and the fail-closed-rather-than-drop behavior is the right default, but it is a real stall vector that did not exist when the DLQ record was a compact body contract. Suggested: truncate `raw_payload` to a bounded size (with a `truncated: true` / original-length marker) before enveloping, or spill oversized payloads elsewhere and keep only a locator. Not a merge blocker.

### P2 — Ticker starts before the immediate first tick
`loop` creates the `time.Ticker` (`scheduler.go:63`) before running the synchronous immediate `s.tick(ctx)` (`scheduler.go:69`). If that first catch-up tick runs longer than `interval`, `t.C` will already have a tick buffered, so the first *periodic* round fires back-to-back with the initial one, skipping the intended rest. Mostly benign (idempotency + lag gate + run-lock absorb it), but starting/resetting the ticker *after* the immediate tick returns would remove the double-poll burst. Optional polish.

### nit — `channel_type` omitempty drops a legitimate `channel_type=0`
`DLQEnvelope.ChannelType` uses `json:"channel_type,omitempty"` (`dlq.go:41`). A row whose channel type is the zero value will omit the field from the forensic record. Low impact — `shard_table` + `source_id` still locate the exact row — but if `0` is a meaningful channel type in this domain, consider dropping `omitempty` for triage fidelity. (`MessageID` is intentionally *not* omitempty, which is correct since it's the Kafka key.)

## 4. Coverage / blind spots
- The base64 encoding of `raw_payload` (standard Go `[]byte` JSON behavior) means **external replay/triage tooling must base64-decode `raw_payload`** to inspect original bytes. That tooling is out of this repo, so I cannot verify it expects base64 — flagging for whoever owns the DLQ consumer side. (This is the encoding facet of the consensus P2 above.)
- The paired octo-deployment change (wires this service into the `search` profile, OFF by default) is in a separate repo and out of scope here; the PR's "merge this first" sequencing note is reasonable given the published image must carry the envelope + first-tick behavior.

---

*Independent model review legs (line-level + long-context cross-file) were absorbed into this review; both completed successfully. No findings rise to a merge blocker — approving with the P2/nit items above as follow-up polish.*

PR #21, opened 2026-06-22
feat(es-indexer): CDC-style writes — full payload projection + payloadRaw + consumer-side visibility fail-closed

yujiawei:
## What

CDC-style write model for the search indexer (root-causes the systemic write/read asymmetry: the reader queries ≥9 payload fields but the indexer only wrote `payload.type` + `payload.text.content`, so image/file names, forward contents, voice/video/richtext were all unsearchable).

- **producer** degrades to shipping the **raw payload verbatim** (`RawPayload`) + metadata.
- **es-indexer consumer** does the full payload projection (ported from the CDC blueprint `ToDoc`, all 8 types: text/image/gif/voice/video/file/mergeForward/richText) + retains `payloadRaw` verbatim (`enabled:false` BLOB → future reader fields need only a local reindex, no re-backfill).
- Collection source stays **poll/SQL** (DB-portable; no binlog/CDC). Only the write shape changes.

## 🔴 P0 — visibility fail-closed moved to the consumer (authoritative)

The visibility ACL (`visibles`/`spaceId`) is now parsed at a **single authoritative landing point**: the consumer `processBatch` pre-check (and the backfill path), via the shared octo-lib `searchmsg.ExtractVisibility`, **fail-closed** — an empty/unparseable `visibles` on a non-encrypted message goes to DLQ (`reason=visibility_untrusted`, `status=0`), it is **never written with empty visibles** (which the reader treats as fail-OPEN). The CDC blueprint's raw `visibles`/`space_id` direct-read (no fail-closed) is deliberately **not** ported. `buildPayloadFromRaw` is pure body projection — it never touches visibility.

Three-branch dispatch in `DocFromMessage` (by `RawPayload` first):
- **A** `len(RawPayload)>0` (non-encrypted new shape, incl. unknown-type group msgs) → projection + consume pre-filled visibility + `payloadRaw`.
- **B** encrypted (`RawPayload=nil` + `RawExcluded`) → skip parse, leave visibles empty (reader fail-closed safe).
- **C** legacy in-flight v2 (no `RawPayload`) → trust contract visibles (transition; backfill rewrites via ExtractVisibility).

## Other safety/correctness

- No `SchemaVersion` bump (consumer strict-equality; a bump would dead-letter all in-flight v2).
- Kafka oversize guard: producer degrades (drop RawPayload → text-only) else truncated DLQ envelope; consumer + producer DLQ envelopes truncate oversized payloads (base64-aware threshold).
[...]

PR #23, opened 2026-06-22
test(harness): gap-table end-to-end regression (CDC-style writes)

### LINKED ISSUE #24: test(harness): machine-checkable gap-table end-to-end regression for CDC-style writes
## Context

PR #21 landed the CDC-style writes work (full payload projection + `payloadRaw` + consumer-side visibility fail-closed). That change touches the entire live ingest path (Kafka → consumer projection → OpenSearch) and the P0 visibility safety gate, but there was no machine-checkable end-to-end regression locking the behavior in.

## Ask

Add an API-only (zero-browser) end-to-end regression that drives the full CDC-style live path against a throwaway local stack (Kafka + OpenSearch-IK) and asserts every row of the plan v2.1 gap table plus the P0 visibility fail-closed safety gate:

- **migrate-forward fail-closed gate** — a stale-mapping index makes es-indexer refuse to start (`AssertLiveMappingCompatible`).
- **gap-table recall** — image/file/mergeForward(from/searchText/timestamp)/voice/video/richText projection + top-level `messageId/messageSeq/channelId/channelType/from/timestamp/spaceId` + snowflake precision + `payloadRaw` retention.
- **P0 visibility security** — valid `visibles` indexed; empty-array / null / non-string `visibles` → DLQ (`visibility_untrusted`, not fail-OPEN); broadcast (no `visibles` key) allowed; encrypted DM (`RawExcluded`) parity.

Throwaway local stack only — never wired into any shared/real environment.

## Acceptance

- `harness/gaptable-e2e.sh` + `harness/gaptable/` drive the full live path and assert all gap-table fields + the fail-closed gate, green on a cold start.
- `golangci-lint run ./...` clean; `go build/vet/test ./...` green.

yujiawei:
## What

Adds a machine-checkable, API-only (zero-browser) end-to-end regression for the
CDC-style writes work (full payload projection + payloadRaw + consumer-side
visibility fail-closed), and fixes one stale pre-CDC assertion in the backfill
harness.

### New: `harness/gaptable-e2e.sh` + `harness/gaptable/`

Drives the full live path against the throwaway local harness stack
(Kafka + OpenSearch-IK) and asserts every gap-table field + the visibility
fail-closed safety gate. **26/26 PASS** on a cold start.

- **STEP0** — stale-mapping fail-closed gate: an index missing the new fields
  makes es-indexer refuse to start (`AssertLiveMappingCompatible`), proving
  `make migrate-forward` is a hard pre-requisite, not optional.
- **STEP1** — `make migrate-forward`: new contract index (payloadRaw
  `enabled:false` + `payload.mergeForward.msgs.{from,timestamp}` +
  `payload.richText.searchText`, `dynamic:strict`) + read-alias switch.
- **STEP2** — es-indexer starts (startup mapping-compat + safety gate OK).
- **STEP3** — reader-DSL recall of every gap-table field: image
  `{caption,name}`, file `{caption,name,extension}`, mergeForward
  `{searchText,from,timestamp}`, voice/video projection, richText searchText,
  top-level `messageId/messageSeq/channelId/channelType/from/timestamp/spaceId`,
  snowflake precision, payloadRaw retention.
- **STEP4** — P0 security: valid `visibles` indexed+populated; empty-array /
  null / non-string `visibles` → DLQ (`visibility_untrusted`, **not fail-OPEN**);
  broadcast (no `visibles` key) allowed; encrypted DM (RawExcluded) parity.

### Fix: `harness/backfill/main.go`

Under CDC-style writes an image row is now PROJECTED (`payload.image.url`), not
`raw_excluded`, so `wantRawExcluded` is 1 (only the encrypted row), not 2. The
backfill existing-data e2e now passes again (doc count, recall, fail-closed
enrichment of stored `visibles` all green).

## How to run

```
./harness/gaptable-e2e.sh            # full live-path gap-table e2e (up→migrate→seed→assert→down)
./harness/run-backfill.sh            # existing-data backfill e2e
```

Throwaway local stack only — never wired into any shared/real environment.

Part of #21

Closes #24

### DIFF
--- a/harness/gaptable/main.go
+++ b/harness/gaptable/main.go
@@ -1,0 +1,149 @@
+// Command gaptable seeds the plan v2.1 gap-table verification vectors into the
+// harness Kafka topic, exercising the Plan B (CDC-style) RawPayload projection +
+// consumer-side visibility fail-closed path end to end.
+//
+// Each message is a searchmsg.Message carrying RawPayload (the raw payload整包),
+// so the es-indexer consumer does: ExtractVisibility pre-check (fail-closed) +
+// buildPayloadFromRaw projection + payloadRaw retention. This is the live path
+// the gap table must be checked against.
+//
+// ISOLATION: connects only to the local harness Kafka (KAFKA_BROKERS, default
+// localhost:19092). Never points at a shared environment.
+package main
+
+import (
+	"context"
+	"encoding/json"
+	"flag"
+	"log"
+	"os"
+	"strings"
+	"time"
+
+	"github.com/Mininglamp-OSS/octo-lib/contract/searchmsg"
+	"github.com/segmentio/kafka-go"
+)
+
+func main() {
+	log.SetFlags(log.LstdFlags | log.LUTC)
+	brokers := flag.String("brokers", envOr("KAFKA_BROKERS", "localhost:19092"), "kafka brokers (csv)")
+	topic := flag.String("topic", envOr("KAFKA_TOPIC", "octo.message.v1"), "body topic")
+	flag.Parse()
+
+	w := &kafka.Writer{
+		Addr:                   kafka.TCP(splitCSV(*brokers)...),
+		Topic:                  *topic,
+		Balancer:               &kafka.Hash{},
+		RequiredAcks:           kafka.RequireAll,
+		Async:                  false,
+		AllowAutoTopicCreation: true,
+	}
+	defer func() { _ = w.Close() }()
+
+	ctx, cancel := context.WithTimeout(context.Background(), 2*time.Minute)
+	defer cancel()
+
+	base := time.Now().Unix()
+	msgs := vectors(base)
+	if err := w.WriteMessages(ctx, msgs...); err != nil {
+		log.Fatalf("produce: %v", err)
+	}
+	log.Printf("gaptable: seeded %d vectors to %s", len(msgs), *topic)
+}
+
+// rawMsg builds a non-encrypted Plan B contract message: RawPayload carries the
+// raw payload整包; the consumer projects + extracts visibility from it.
+func rawMsg(id, channelID string, channelType int, from string, seq uint64, ts int64, rawPayload string) kafka.Message {
+	m := searchmsg.Message{
+		SchemaVersion: searchmsg.SchemaVersion,
+		MessageID:     id,
+		ChannelID:     channelID,
+		ChannelType:   channelType,
+		FromUID:       from,
+		MessageSeq:    seq,
+		MsgTimestamp:  ts,
+		CreatedAt:     ts,
+		RawPayload:    json.RawMessage(rawPayload),
+		Source:        searchmsg.SourceETLMessageTable,
+	}
+	b, err := json.Marshal(m)
+	if err != nil {
+		log.Fatalf("marshal %s: %v", id, err)
+	}
+	return kafka.Message{Key: []byte(id), Value: b}
+}
+
+func vectors(base int64) []kafka.Message {
+	var out []kafka.Message
+	// ---- Gap-table positive recall vectors (each via RawPayload projection) ----
+	// G1 image: payload.image.{caption,name}
+	out = append(out, rawMsg("3000000000000000001", "g_gap", 2, "u_img", 11, base,
+		`{"type":2,"name":"季度合同.png","caption":"图片说明文字甲","url":"https://x/a.png","width":800,"height":600}`))
+	// G2 file: payload.file.{caption,name,extension}
+	out = append(out, rawMsg("3000000000000000002", "g_gap", 2, "u_file", 12, base,
+		`{"type":8,"name":"年度报告.pdf","caption":"文件说明文字乙","url":"https://x/r.pdf","size":123456,"extension":"pdf"}`))
+	// G3 mergeForward: payload.mergeForward.msgs.{searchText,from,timestamp}
+	out = append(out, rawMsg("3000000000000000003", "g_gap", 2, "u_mf", 13, base,
+		`{"type":11,"msgs":[{"message_id":"9007199254740993","from_uid":"u_inner_丙","timestamp":1700000123,"payload":{"type":1,"content":"转发卡内文字丙"}},{"message_id":"9007199254740994","from_uid":"u_inner_丁","timestamp":1700000456,"payload":{"type":8,"name":"内层文件丁.doc"}}]}`))
+	// G4 voice (projection present)
+	out = append(out, rawMsg("3000000000000000004", "g_gap", 2, "u_voice", 14, base,
+		`{"type":4,"url":"https://x/v.mp3"}`))
+	// G5 video (projection present)
+	out = append(out, rawMsg("3000000000000000005", "g_gap", 2, "u_video", 15, base,
+		`{"type":5,"url":"https://x/v.mp4","cover":"https://x/c.jpg","width":1920,"height":1080,"second":42}`))
+	// G6 richtext: payload.richText.searchText
+	out = append(out, rawMsg("3000000000000000006", "g_gap", 2, "u_rich", 16, base,
+		`{"type":14,"content":[{"type":"text","text":"富文本正文戊"},{"type":"image","name":"富文本图片己.png"}]}`))
+	// G7 text top-level fields anchor (messageId/messageSeq/channelId/channelType/from/timestamp/spaceId)
+	out = append(out, rawMsg("3000000000000000007", "g_gap", 2, "u_top", 17, base,
+		`{"type":1,"content":"顶层字段锚定庚","space_id":"space_top"}`))
+
+	// ---- P0 visibility fail-closed vectors (group, non-encrypted) ----
+	// V1 valid visibles → indexed with visibles populated
+	out = append(out, rawMsg("3000000000000000010", "g_acl", 2, "u_adm", 20, base,
+		`{"type":99,"content":"群管可见系统消息","visibles":["u_admin1","u_admin2"]}`))
+	// V2 empty-array visibles → fail-closed → DLQ (NOT indexed, NOT fail-OPEN)
+	out = append(out, rawMsg("3000000000000000011", "g_acl", 2, "u_adm", 21, base,
+		`{"type":99,"content":"valid-but-empty 必须进DLQ","visibles":[]}`))
+	// V3 null visibles → fail-closed → DLQ
+	out = append(out, rawMsg("3000000000000000012", "g_acl", 2, "u_adm", 22, base,
+		`{"type":99,"content":"null visibles 必须进DLQ","visibles":null}`))
+	// V4 visibles all non-string → fail-closed → DLQ
+	out = append(out, rawMsg("3000000000000000013", "g_acl", 2, "u_adm", 23, base,
+		`{"type":99,"content":"全非字符串 visibles 必须进DLQ","visibles":[123,456]}`))
+	// V5 broadcast (no visibles key) → allowed, visibles empty (normal group chat)
+	out = append(out, rawMsg("3000000000000000014", "g_acl", 2, "u_member", 24, base,
+		`{"type":1,"content":"普通群聊广播放行辛"}`))
+
+	// ---- Encrypted DM branch: RawPayload nil + RawExcluded=true → indexed, visibles empty (parity) ----
+	encM := searchmsg.Message{
+		SchemaVersion: searchmsg.SchemaVersion,
+		MessageID:     "3000000000000000020",
+		ChannelID:     "u_x@u_y", ChannelType: 1, FromUID: "u_x",
+		MessageSeq: 30, MsgTimestamp: base, CreatedAt: base,
+		RawExcluded: true, Content: nil,
+		Source: searchmsg.SourceETLMessageTable,
+	}
+	eb, _ := json.Marshal(encM)
+	out = append(out, kafka.Message{Key: []byte(encM.MessageID), Value: eb})
+
+	return out
+}
+
+func envOr(k, def string) string {
+	if v := os.Getenv(k); v != "" {
+		return v
+	}
+	return def
+}
+
+func splitCSV(s string) []string {
+	parts := strings.Split(s, ",")
+	out := make([]string, 0, len(parts))
+	for _, p := range parts {
+		if t := strings.TrimSpace(p); t != "" {
+			out = append(out, t)
+		}
+	}
+	return out
+}
--- a/harness/gaptable/main.go
+++ b/harness/gaptable/main.go
@@ -38,7 +38,11 @@ func main() {
 		Async:                  false,
 		AllowAutoTopicCreation: true,
 	}
-	defer func() { _ = w.Close() }()
+	defer func() {
+		if cerr := w.Close(); cerr != nil {
+			log.Printf("gaptable: writer close error: %v", cerr)
+		}
+	}()
 
 	ctx, cancel := context.WithTimeout(context.Background(), 2*time.Minute)
 	defer cancel()
@@ -124,7 +128,10 @@ func vectors(base int64) []kafka.Message {
 		RawExcluded: true, Content: nil,
 		Source: searchmsg.SourceETLMessageTable,
 	}
-	eb, _ := json.Marshal(encM)
+	eb, err := json.Marshal(encM)
+	if err != nil {
+		log.Fatalf("marshal %s: %v", encM.MessageID, err)
+	}
 	out = append(out, kafka.Message{Key: []byte(encM.MessageID), Value: eb})
 
 	return out
--- a/harness/backfill/main.go
+++ b/harness/backfill/main.go
@@ -29,14 +29,19 @@ import (
 // Seeded suite shape (kept tiny + explicit so the gate arithmetic is obvious):
 //   - 3 normal text rows (incl. 中文)            → indexed, content present
 //   - 1 Signal-encrypted row (signal column)     → raw_excluded, still 1 ES doc
-//   - 1 non-text (image) row                     → raw_excluded, still 1 ES doc
+//   - 1 non-text (image) row                     → PROJECTED (payload.image.url),
+//     NOT raw_excluded (Plan B / CDC-style: docFromRow now projects every typed
+//     payload from the raw payload整包 and recomputes RawExcluded from whether a
+//     typed sub-object was produced; an image row yields payload.image so it is a
+//     normal projected doc, not a raw-excluded one).
 //   - 1 bad-JSON non-encrypted row (real anomaly)→ DLQ spill, NOT in ES
 //
-// So: source_rows=6, expected ES docs=5 (6 - 1 DLQ), raw_excluded=2.
+// So: source_rows=6, expected ES docs=5 (6 - 1 DLQ), raw_excluded=1 (only the
+// Signal-encrypted row; the image row is projected post-Plan-B).
 const (
 	wantSource      = 6
 	wantESDocs      = 5
-	wantRawExcluded = 2
+	wantRawExcluded = 1
 	wantDLQ         = 1
 )
 
@@ -182,7 +187,7 @@ func verify(ctx context.Context, esURL, index string, base int64) error {
 		log.Printf("FAIL: raw_excluded=%d want %d", rawEx, wantRawExcluded)
 		fail = true
 	} else {
-		log.Printf("PASS: raw_excluded=%d (Signal + non-text, still occupy ES docs)", rawEx)
+		log.Printf("PASS: raw_excluded=%d (only the Signal-encrypted row; image row is projected post-Plan-B)", rawEx)
 	}
 	// Positively assert the bad-JSON row is NOT in ES (it went to DLQ spill).
 	badPresent, err := count(ctx, client, index, idsQuery("3000000000000000006"))
--- a/harness/gaptable-e2e.sh
+++ b/harness/gaptable-e2e.sh
@@ -1,0 +1,168 @@
+#!/usr/bin/env bash
+# Plan v2.1 gap-table end-to-end regression (API-only, zero browser).
+#
+# Drives the FULL Plan B (CDC-style writes) live path against a throwaway local
+# stack (Kafka + OpenSearch-IK from harness/docker-compose.yml) and machine-checks
+# every row of the plan v2.1 gap table + the P0 visibility fail-closed safety闸:
+#
+#   0. P0 migrate-forward fail-closed gate: a stale-mapping index makes es-indexer
+#      REFUSE to start (AssertLiveMappingCompatible) → proves `make migrate-forward`
+#      is a hard pre-req, not optional.
+#   1. make migrate-forward: new contract index (payloadRaw enabled:false +
+#      image/file/mergeForward.{from,timestamp}/voice/video/richText, dynamic:strict)
+#      + read-alias switch.
+#   2. es-indexer live: Kafka octo.message.v1 -> consumer (RawPayload projection +
+#      ExtractVisibility fail-closed pre-check) -> OpenSearch.
+#   3. Gap-table reader-DSL recall: image/file/mergeForward(from/searchText/ts)/
+#      voice/video/richText + top-level messageId/messageSeq/channelId/channelType/
+#      from/timestamp/spaceId + snowflake precision + payloadRaw retention.
+#   4. P0 security: valid visibles indexed; empty/null/non-string visibles -> DLQ
+#      (visibility_untrusted, NOT fail-OPEN); broadcast (no visibles key) allowed;
+#      encrypted DM (RawExcluded) parity (no payload, empty visibles).
+#
+# ISOLATION: throwaway local stack ONLY (harness project). Never wired into any
+# shared/real environment. The standalone backfill existing-data path lives in
+# harness/run-backfill.sh.
+#
+# Usage:
+#   ./harness/gaptable-e2e.sh           # full run: up -> migrate -> seed -> assert -> down
+#   KEEP_UP=1 ./harness/gaptable-e2e.sh # leave the stack up after asserting
+set -euo pipefail
+
+HERE="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+ROOT="$(cd "$HERE/.." && pwd)"
+cd "$ROOT"
+
+ES_URL="${ES_URL:-http://localhost:19200}"
+KAFKA_EXTERNAL="${KAFKA_EXTERNAL:-localhost:19092}"
+NEW_INDEX="${NEW_INDEX:-wukongim-messages-000001}"
+ALIAS="${ALIAS:-wukongim-messages-read}"
+OLD_INDEX="${OLD_INDEX:-octo-message}"
+INDEXER_LOG=/tmp/octo-gaptable-indexer.log
+INDEXER_PID_FILE=/tmp/octo-gaptable-indexer.pid
+
+# BIN_DIR (optional): a directory of pre-built binaries (es-indexer, gaptable).
+# When set, the script runs those instead of `go run` (handy in CI / sandboxes
+# where `go run`'s child-process signal handling is awkward). When empty the
+# script self-builds them once into a temp dir.
+BIN_DIR="${BIN_DIR:-}"
+if [ -z "$BIN_DIR" ]; then
+  BIN_DIR="$(mktemp -d)"
+  echo "[gaptable] building binaries into $BIN_DIR ..."
+  go build -o "$BIN_DIR/es-indexer" ./cmd/es-indexer
+  go build -o "$BIN_DIR/gaptable" ./harness/gaptable
+fi
+ES_INDEXER_BIN="$BIN_DIR/es-indexer"
+GAPTABLE_BIN="$BIN_DIR/gaptable"
+
+PASS=0; FAIL=0
+ok(){ echo "PASS | $1"; PASS=$((PASS+1)); }
+no(){ echo "FAIL | $1"; FAIL=$((FAIL+1)); }
+
+jqid(){ python3 -c "import json,sys;r=json.load(sys.stdin);print(','.join(h['_id'] for h in r['hits']['hits']))"; }
+
+recall(){ # label dsl want
+  local got; got=$(curl -s "$ES_URL/$ALIAS/_search" -H 'Content-Type: application/json' -d "$2" | jqid)
+  if echo ",$got," | grep -q ",$3,"; then ok "$1 -> $3"; else no "$1 -> want $3 got [$got]"; fi
+}
+src(){ curl -s "$ES_URL/$NEW_INDEX/_doc/$1"; } # fetch _doc json
+
+stop_indexer(){ [ -f "$INDEXER_PID_FILE" ] && { kill "$(cat "$INDEXER_PID_FILE")" 2>/dev/null||true; rm -f "$INDEXER_PID_FILE"; }; }
+down(){ echo "[gaptable] tearing down..."; stop_indexer; ( cd "$HERE" && docker compose down -v --remove-orphans ) || true; }
+
+echo "[gaptable] === version assert ==="
+git rev-parse HEAD
+grep octo-lib go.mod
+
+echo "[gaptable] === bring up throwaway Kafka + OpenSearch(IK) ==="
+( cd "$HERE" && docker compose up -d --build )
+for _ in $(seq 1 60); do curl -fs "$ES_URL/_cluster/health" >/dev/null 2>&1 && break; sleep 3; done
+curl -fs "$ES_URL/_cluster/health?wait_for_status=yellow&timeout=60s" >/dev/null
+for _ in $(seq 1 60); do docker exec octo-harness-kafka /opt/kafka/bin/kafka-broker-api-versions.sh --bootstrap-server localhost:9092 >/dev/null 2>&1 && break; sleep 3; done
+for t in octo.message.v1 octo.message.v1.dlq; do
+  docker exec octo-harness-kafka /opt/kafka/bin/kafka-topics.sh --bootstrap-server localhost:9092 --create --if-not-exists --topic "$t" --partitions 1 --replication-factor 1 >/dev/null 2>&1 || true
+done
+[ "${KEEP_UP:-0}" = "1" ] || trap down EXIT
+
+echo "[gaptable] === STEP 0: P0 migrate-forward fail-closed gate ==="
+curl -s -XDELETE "$ES_URL/$OLD_INDEX" >/dev/null 2>&1 || true
+curl -s -XDELETE "$ES_URL/$NEW_INDEX" >/dev/null 2>&1 || true
+STALE='{"mappings":{"dynamic":"strict","properties":{"messageId":{"type":"long"},"payload":{"type":"object","properties":{"type":{"type":"integer"},"text":{"type":"object","properties":{"content":{"type":"text"}}}}}}}}'
+curl -s -XPUT "$ES_URL/$OLD_INDEX" -H 'Content-Type: application/json' -d "$STALE" >/dev/null
+if env ES_INDEXER_ENABLED=true KAFKA_BROKERS="$KAFKA_EXTERNAL" KAFKA_TOPIC=octo.message.v1 KAFKA_DLQ_TOPIC=octo.message.v1.dlq KAFKA_GROUP_ID=octo-gaptable-gate ES_ADDRESSES="$ES_URL" ES_INDEX="$OLD_INDEX" INDEXER_DLQ_SPILL_DIR=/tmp/octo-e2e-gate-spill timeout 30 "$ES_INDEXER_BIN" >/tmp/octo-gate.log 2>&1; then
+  no "STEP0 stale-mapping gate (indexer should have refused to start)"
+else
+  if grep -q 'mapping-compat FAILED' /tmp/octo-gate.log && grep -q 'make migrate-forward' /tmp/octo-gate.log; then
+    ok "STEP0 stale-mapping gate: es-indexer fail-closed, demands make migrate-forward"
+  else
+    no "STEP0 stale-mapping gate: indexer exited but not with mapping-compat failure"; cat /tmp/octo-gate.log
+  fi
+fi
+
+echo "[gaptable] === STEP 1: make migrate-forward ==="
+ES="$ES_URL" ALIAS="$ALIAS" OLD_INDEX="$OLD_INDEX" NEW_INDEX="$NEW_INDEX" STEP=1 ./scripts/forward-migrate.sh >/dev/null 2>&1
+ES="$ES_URL" ALIAS="$ALIAS" OLD_INDEX="$OLD_INDEX" NEW_INDEX="$NEW_INDEX" STEP=2 ./scripts/forward-migrate.sh >/dev/null 2>&1
+ES="$ES_URL" ALIAS="$ALIAS" OLD_INDEX="$OLD_INDEX" NEW_INDEX="$NEW_INDEX" STEP=3 ./scripts/forward-migrate.sh >/dev/null 2>&1
+MF=$(curl -s "$ES_URL/$NEW_INDEX/_mapping")
+echo "$MF" | python3 -c "import json,sys;m=json.load(sys.stdin)['$NEW_INDEX']['mappings']['properties'];assert m['payloadRaw']['enabled'] is False;mf=m['payload']['properties']['mergeForward']['properties']['msgs']['properties'];assert 'from' in mf and 'timestamp' in mf;assert 'richText' in m['payload']['properties']" \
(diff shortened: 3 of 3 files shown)


OctoBoooot (commented):
## Review: test(harness): gap-table end-to-end regression (CDC-style writes) (#23)

**Verdict: Comment** (per harness convention `APPROVE` is also fine if `check-sprint` is unblocked).
CI is red — `check-sprint / Validate Sprint field on Project Board` (process gate, not code). Build/Test/Vet/CodeQL/osv all still pending; no code-level failures.

Risk tier: **normal** (test harness only, no auth/queue/infra code).

### Notes
- `harness/backfill/main.go:32-46` — the `wantRawExcluded` bump from 2 → 1 correctly reflects the post-Plan-B behavior verified in #21: image rows now produce a typed `payload.image.*` projection so `RawExcluded` is recomputed as `false`, leaving only the Signal-encrypted row in the raw-excluded bucket. The updated PASS-line message accurately explains why.
- `harness/gaptable-e2e.sh` — solid end-to-end gate: STEP 0 (stale-mapping fail-closed, the gate I most wanted machine-checked) → STEP 1 (migrate-forward) → STEP 2 (indexer start) → STEP 3 (gap-table reader-DSL recall, including the snowflake `9007199254740993 > 2^53` precision check at line 189) → STEP 4 (visibility fail-closed with `valid-but-empty` / `null` / `non-string` `visibles` all going to DLQ as `visibility_untrusted`). The DLQ-offset fence (`DLQ_START` capture before seeding, slice from there) at lines 167-205 cleanly handles re-runs against a `KEEP_UP=1` stack — that's exactly the right design for an idempotent harness.
- `harness/gaptable/main.go` vectors hit every cell of the gap table I'd want regression-tested for the CDC-style writes work (image caption/name, file caption/name/extension, mergeForward.{searchText,from,timestamp} + a snowflake-id child, voice/video projections, richText, top-level fields, ACL fail-closed edge cases, encrypted-DM parity).

### Minor
- `harness/gaptable-e2e.sh:184-186` — the top-level field assertion only checks `timestamp > 0`. Since `base := time.Now().Unix()` is the produced value, a tighter `>= test_start_unix` would catch a future regression that swaps timestamp sources. Not a blocker — current check still catches "field absent" / "field zero".

### Praise
- The STEP 0 stale-mapping gate is the right shape — it asserts the *negative* path (es-indexer MUST refuse to start) by checking the specific failure mode in logs (`mapping-compat FAILED` + `make migrate-forward`), not just exit code. Future readers who edit `AssertLiveMappingCompatible` will trip exactly this test rather than silently degrading the fail-closed contract.
- The DLQ-fencing pattern with offset slice + `kafka-console-consumer` reason-grep gives precise "this run's records" semantics without resetting state. Worth replicating in other harnesses.

### Out of scope
- `check-sprint` failure is a project-management gate (Sprint field on the project board), not a code defect. Same nuisance as #21 — needs the Sprint set to clear.
- Coverage-sniff would flag `harness/backfill/main.go` and `harness/gaptable/main.go` as having no `_test.go` — but they ARE the test harness; no tests-for-tests expected. Dropping.

Jerry-Xin (changes requested):
Adds relevant local E2E coverage for the indexer’s CDC-style write path and updates the backfill harness expectation consistently with the current projection behavior.

💬 Non-blocking

🟡 Warning: [harness/gaptable-e2e.sh:116](/Users/mlamp/.openclaw/tmp/pr-review-o48xyje8/repo/harness/gaptable-e2e.sh:116) and [harness/gaptable-e2e.sh:128](/Users/mlamp/.openclaw/tmp/pr-review-o48xyje8/repo/harness/gaptable-e2e.sh:128) rely on fixed sleeps for indexer readiness and ingestion completion. This can be flaky on slow Docker hosts or CI. Polling for the expected indexed documents and DLQ offsets would make the regression more deterministic.

🟡 Warning: [harness/gaptable-e2e.sh:65](/Users/mlamp/.openclaw/tmp/pr-review-o48xyje8/repo/harness/gaptable-e2e.sh:65) can exit the whole script under `set -euo pipefail` if OpenSearch returns an error/non-search JSON, bypassing the script’s `FAIL` accounting and final summary. Consider making `recall` handle parse/search errors explicitly and call `no`.

🔵 Suggestion: [harness/gaptable/main.go:127](/Users/mlamp/.openclaw/tmp/pr-review-o48xyje8/repo/harness/gaptable/main.go:127) ignores the `json.Marshal` error for the encrypted vector. It is unlikely to fail with the current struct, but checking it keeps the harness consistent with `rawMsg`.

✅ Highlights

- The PR is in scope for this repository: it extends the local test harness around indexer, migration, OpenSearch mapping, and backfill behavior.
- The new E2E script covers useful security-critical cases: stale mapping fail-closed startup, visibility fail-closed DLQ routing, broadcast allowance, and encrypted DM parity.
- DLQ offset fencing is a good choice for repeatable runs on kept-up stacks.
- `go test ./...` passes.

mochashanyao (dismissed):
[Octo-Q · automated review]

---

# octo-search-indexer PR#23 Review Report
**PR:** https://github.com/Mininglamp-OSS/octo-search-indexer/pull/23
**Head SHA:** `af9e3081579568748dfd69a6a66fe168be4253ba`
**Title:** test(harness): gap-table end-to-end regression (CDC-style writes)
**Scope:** 3 files, +326/-4 (test harness only — zero production code changes)

---

## 1. Verification Summary

| Item | Status | Evidence |
|------|--------|----------|
| `wantRawExcluded` 2→1 fix | ✅ Correct | `doc_enrich.go:49` sets `msg.RawPayload` for non-encrypted rows → `DocFromMessage` branch A (`doc.go:163`) → `buildPayloadFromRaw` projects image (type=2) → `projected=true` → `RawExcluded=false`. Only Signal-encrypted row remains raw_excluded. |
| STEP0 stale-mapping gate | ✅ Correct | `mapping_compat.go:28-33` requires `payload.mergeForward.msgs.{from,timestamp}` + `payload.richText.searchText`; stale mapping lacks these → `AssertLiveMappingCompatible` returns error → indexer refuses to start. |
| STEP1 migrate-forward | ✅ Correct | `scripts/forward-migrate.sh` STEP 1/2/3 creates new index from embedded mapping (`mapping/octo-message.json`), reindexes, switches alias. Test verifies mapping fields + alias. |
| STEP3 recall (all 10 queries) | ✅ Correct | Each DSL query targets fields produced by `buildPayloadFromRaw` (`buildraw.go:79-155`): image.caption/name, file.caption/name/extension, mergeForward.msgs.searchText/from/timestamp, richText.searchText. Field mappings verified against production code. |
| Snowflake precision | ✅ Correct | `decodeObjectUseNumber` (`buildraw.go:172`) uses `json.Decoder.UseNumber()` preventing float64 truncation of >2^53 IDs. Test asserts `9007199254740993 in ids`. |
| payloadRaw retention | ✅ Correct | `buildPayloadFromRaw` returns `rawForStore = rawPayload` for valid JSON objects (`buildraw.go:67`). Test asserts `payloadRaw.width==800`. |
| STEP4 visibility fail-closed | ✅ Correct | `ExtractVisibility` (`octo-lib/visibility.go`): empty array → zero valid UIDs → fail-closed; null → fail-closed; non-string elements → zero valid UIDs → fail-closed; absent key → broadcast (nil error, nil visibles). All 5 vectors match. |
| DLQ fencing | ✅ Correct | Captures DLQ log-end offset before seeding, counts only new records. Cross-checks `visibility_untrusted` reason via console-consumer. |
| Encrypted DM parity (V20) | ✅ Correct | `RawExcluded:true, RawPayload:nil` → `DocFromMessage` branch B (`doc.go:172`) → no projection, `RawExcluded=true`, `Payload=nil`, `Visibles=nil`. |

## 2. Findings

### No P0 or P1 findings.

### P2 — Minor test-harness robustness notes (non-blocking)

**P2-1: Hard-coded sleeps for indexer startup / message processing**
- `sleep 12` (STEP2, indexer startup) and `sleep 8` (post-seed, message processing) are fixed waits. Under heavy CI load these could be insufficient, causing flaky failures.
- The script already has retry loops for Kafka/OpenSearch startup (good), but the indexer and processing waits are fixed.
- **Suggestion:** Poll the indexer log for `es-indexer running` in a loop (like the Kafka/OpenSearch waits), and poll ES for expected doc count after seeding.
- **Severity:** P2/nit — common pattern in test harnesses, not a correctness issue on the PR's stated scope (throwaway local stack).

**P2-2: `recall` function uses substring match**
- `grep -q ",$3,"` does substring matching on comma-delimited IDs. With 19-digit snowflake IDs the collision probability is negligible, but it's not strict equality.
- **Severity:** P2/nit — acceptable for test harness.

## 3. Suggestions

1. Replace `sleep 12` / `sleep 8` with poll loops (check indexer log / ES doc count) for CI robustness.
2. Consider adding a negative recall test (assert a DLQ'd vector is NOT findable via the read alias) to complement the `found=False` check in STEP4.

## 4. Additional Observations

- **Backfill stats counter discrepancy (pre-existing, not introduced by this PR):** `extractMessage` in `extract.go` classifies non-text types as `outcomeRawExcluded`, but `docFromRow` overrides `RawExcluded` in the ES doc via branch A projection. The runner's stats counter (`s.RawExcluded++`) uses the `extractMessage` outcome, not the final doc's `RawExcluded`. This means backfill stats may over-count `RawExcluded` for projected non-text types. This is a stats-only cosmetic issue (doesn't affect ES docs) and is pre-existing. Not blocking this PR.

## 5. Data-Flow Tracing

|---|---|---|
| `wantRawExcluded=1` | `doc_enrich.go:49` → `doc.go:163-170` (branch A, `projected=true` for image) | ✅ Yes — image row projected, only Signal-encrypted remains raw_excluded |
| `payload.image.caption/name` | `buildraw.go:91-101` (ImagePayload projection) | ✅ Yes — raw `caption`/`name` strings projected |
| `payload.file.extension` | `buildraw.go:131-143` (FilePayload projection) | ✅ Yes — raw `extension` string projected |
| `payload.mergeForward.msgs.from` | `buildraw.go:196` (`from_uid` → `From`) | ✅ Yes — snake_case `from_uid` mapped to reader-contract `from` |
| `payload.mergeForward.msgs.searchText` | `buildraw.go:200-204` (sub-payload type dispatch → `buildSearchText`) | ✅ Yes — inner text content + file name concatenated |
| `payload.richText.searchText` | `buildraw.go:228-244` (`richTextSearchText`: plain + image/file block names) | ✅ Yes — text blocks + image names concatenated |
| `payloadRaw.width` | `buildraw.go:67` (`rawForStore = rawPayload`) | ✅ Yes — raw JSON object stored as-is |
| `visibles` (V1) | `consumer.go:128-134` (ExtractVisibility → backfill `msg.Visibles`) | ✅ Yes — valid string array propagated to doc |
| DLQ routing (V2/V3/V4) | `consumer.go:131` (ExtractVisibility err → `visibilityInvalid`) → `decision.go:228` (`reasonVisibilityUntrusted`) | ✅ Yes — fail-closed → DLQ with `visibility_untrusted` reason |
| Snowflake `9007199254740993` | `buildraw.go:172` (`UseNumber`) → `extractInt64` (`json.Number.Int64()`) | ✅ Yes — no float64 truncation |

- **C1 Dual-path parity:** N/A — test-only PR, no add/remove or create/delete symmetry paths.
- **C2 Control-flow ordering / nested reuse:** N/A — test exercises existing production code sequentially (STEP0→4), no new production logic.
- **C3 Authorization boundary:** N/A — harness connects only to local throwaway stack.
- **C4 Authorization lifecycle:** N/A — no auth changes.
- **C5 Build ≠ runtime:** Clear — binaries built explicitly (`go build`), runtime is controlled Docker stack, no relative URL / origin issues.
- **C6 Governance / policy docs:** N/A — no governance docs changed.

## 7. Cross-round Blocker Recheck (R6)

---

**[Octo-Q] verdict: APPROVE** — No P0/P1 findings. Well-structured test harness with correct assertions verified against production code data-flow. Adds valuable end-to-end coverage for Plan B (CDC-style writes) including the P0 visibility fail-closed safety gate. Minor P2 nits on sleep-based timing are non-blocking.

yujiawei (commented):
# Code Review — PR #23 (octo-search-indexer)

**Verdict: APPROVED** (no blocking findings). Reviewed against head `af9e308`; merge-base `main`. Scope: a test harness — a new gap-table e2e shell driver (`harness/gaptable-e2e.sh`), a new Kafka seeder (`harness/gaptable/main.go`), and a single constant change in `harness/backfill/main.go`. No production code is modified.

## 1. Verification

- ✅ **Build / vet / format / tests**: `go build ./...` and `go vet ./harness/...` clean; `gofmt -l` empty; `shellcheck -S warning` clean (`bash -n` OK); `go test ./internal/consumer/... ./internal/esindex/... ./internal/backfill/...` all green.
- ✅ **`wantRawExcluded` 2 → 1 is correct.** The seeded image row `{"type":2,"url":...}` (`harness/backfill/main.go:114`, `signal=0`) is non-encrypted, so `docFromRow` (`internal/backfill/doc_enrich.go:35`) runs `ExtractVisibility` (no `visibles` key → broadcast, no error), sets `RawPayload`, and `DocFromMessage` enters branch A (`internal/esindex/doc.go:192`). `buildPayloadFromRaw` unconditionally produces `payload.Image` for `type=2` (`internal/esindex/buildraw.go:81-98`) → `projected=true` → `d.RawExcluded=!projected=false` (`doc.go:203`). Only the Signal-encrypted row stays `rawExcluded=true` (branch B), so the ES `rawExcluded:true` doc count is exactly 1. Arithmetic for the 6-row suite (3 text + 1 signal + 1 image + 1 bad-JSON→DLQ) holds: source=6, ES docs=5, rawExcluded=1, DLQ=1.
- ✅ **Visibility fail-closed assertions match the real contract.** Seed rows 11 `[]`, 12 `null`, 13 `[123,456]` all return an error from `ExtractVisibility` (empty-array / null / zero-valid-UID branches in `contract/searchmsg/visibility.go`), so the consumer pre-check (`internal/consumer/consumer.go:146`) marks them `visibilityInvalid` and routes them to DLQ as `visibility_untrusted` (`consumer.go:304`), excluded from the bulk. Row 10 (valid visibles) indexes populated; row 14 (no key) passes as broadcast; encrypted row 020 (`RawPayload=nil`, `RawExcluded=true`) skips the pre-check and indexes via branch B with empty visibles. All match the STEP 4 assertions.
- ✅ **All grepped strings are real emitted output**: `mapping-compat FAILED` + `make migrate-forward` (`internal/esindex/mapping_compat.go:91-92`), `es-indexer running` (`cmd/es-indexer/main.go:67`), `exited with error` (`main.go:42`), `visibility_untrusted` (`consumer.go:304`). The STEP 0 fail-closed gate is sound: a stale mapping genuinely lacks the required field paths, the assertion error propagates to `os.Exit(1)`, and the migrate-forward `STEP=1/2/3` interface exists and creates `payloadRaw(enabled:false)` + `mergeForward.{from,timestamp}` + `richText`.

This is a thorough, well-isolated harness that machine-checks both the gap-table projection and the P0 visibility safety gate end to end. The fail-closed assertions are notably robust: each "must-DLQ" row is checked **both** for absence from the index **and** via a fenced positive DLQ count of exactly 3 `visibility_untrusted` records — so a single regression cannot silently false-pass.

## 2. Findings (all P2 / nit — non-blocking)

None of the below block merge; they are test-robustness hardening for CI portability and false-pass margin.

- **P2 — Timing via fixed `sleep` instead of poll-until-ready** (`harness/gaptable-e2e.sh:117,128`). `sleep 12` / `sleep 8` for indexer startup and Kafka consumption will be flaky on loaded CI runners. Prefer polling the indexer log marker / ES doc count with a deadline. (This causes false *negatives* far more than false positives, but it is the main CI-stability risk.)
- **P2 — `KEEP_UP=1` reuse / cleanup** (`harness/gaptable-e2e.sh:85,114`). With `KEEP_UP=1` the `trap down EXIT` is skipped, leaving the `nohup` es-indexer running and the Kafka consumer-group offsets uncleared; a reused stack may not re-exercise the same path and the detached process can collide with the next run. Consider always stopping the indexer on exit even when keeping the stack.
- **P2 — DLQ reason cross-check is not pinned to the 3 expected IDs and consumes only partition 0** (`harness/gaptable-e2e.sh:159-164`). The fence + `DLQ_NEW==3 && DLQ_VU==3` is good, but it doesn't assert *which* three IDs landed, and the offset sum spans all partitions while the consumer reads `--partition 0` (safe only because the topic is created with `--partitions 1`). Tying the count to the expected message IDs would close the residual false-pass margin. (Mitigated today by the independent per-ID absence checks at lines 152-155.)
- **P2 — Non-hermetic host deps**: fixed ports `19200`/`19092` and shared `/tmp` paths (`gaptable-e2e.sh:36-37,41-42`) can collide across concurrent runs on one host; `python3` is an unguarded hard dependency for every assertion. Fine for local dev; worth isolating (`mktemp -d`, per-run ports) before wiring into shared CI.
- **nit — Python `assert` for the mapping check** (`gaptable-e2e.sh:107`) is silently disabled under `PYTHONOPTIMIZE`/`-O`; **alias check is substring-only** (`:110`) so a multi-index alias could pass. Use explicit `sys.exit(1)` and an exact alias-set check.

## 3. Suggestions

- Replace the two fixed `sleep`s with bounded poll loops (the script already has the `for _ in $(seq ...)` pattern for cluster health — reuse it for indexer-ready and doc-count).
- Pin the DLQ reason cross-check to the expected `{11,12,13}` IDs in addition to the count.

## 4. Additional notes

- **Coverage the harness claims but only partially exercises** (for follow-up, not this PR): `ExtractVisibility` is tested for empty/null/all-non-string visibles, but not the mixed valid+invalid case (some entries valid, some not) — a claimed fail-closed dimension that isn't seeded.
- The `go.sum` changes were not line-audited.
- These observations are static-analysis based; the timing budgets may well pass reliably on the author's machine — which is exactly why the CI-load false-pass margin is the thing to watch.

harness/gaptable/main.go
```diff
@@ -0,0 +1,149 @@
...
+
+	w := &kafka.Writer{
+		Addr:                   kafka.TCP(splitCSV(*brokers)...),
+		Topic:                  *topic,
+		Balancer:               &kafka.Hash{},
+		RequiredAcks:           kafka.RequireAll,
+		Async:                  false,
+		AllowAutoTopicCreation: true,
+	}
+	defer func() { _ = w.Close() }()
```
OctoBoooot: **[blocker]** Lint CI FAILURE on this exact line. Repo's `.golangci.yml` sets `errcheck.check-blank: true`, so `_ = ...` doesn't silence errcheck — log says: `Error return value of w.Close is not checked (errcheck)`. Fix: replace with `defer func() { if err := w.Close(); err != nil { log.Printf("close kafka writer: %v", err) } }()`. Without this, the harness's own CI stays red and the regression suite can't merge.

harness/gaptable/main.go
```diff
@@ -0,0 +1,149 @@
...
+	// ---- Encrypted DM branch: RawPayload nil + RawExcluded=true → indexed, visibles empty (parity) ----
+	encM := searchmsg.Message{
+		SchemaVersion: searchmsg.SchemaVersion,
+		MessageID:     "3000000000000000020",
+		ChannelID:     "u_x@u_y", ChannelType: 1, FromUID: "u_x",
+		MessageSeq: 30, MsgTimestamp: base, CreatedAt: base,
+		RawExcluded: true, Content: nil,
+		Source: searchmsg.SourceETLMessageTable,
+	}
+	eb, _ := json.Marshal(encM)
```
OctoBoooot: **[blocker]** Same Lint CI failure: `Error return value of json.Marshal is not checked (errcheck)`. Fix: `eb, err := json.Marshal(encM); if err != nil { log.Fatalf("marshal encrypted vector: %v", err) }` — `json.Marshal` on `searchmsg.Message` shouldn't fail in practice, but errcheck under `check-blank: true` requires an explicit handler (matching the pattern at line 73 for `rawMsg`).

OctoBoooot (changes requested):
## Review: test(harness): gap-table end-to-end regression (#23) — follow-up at af9e3081

**Verdict: Request changes** (upgrading from my earlier COMMENT).
My prior review was posted while `Lint` was still pending; I should have waited or pre-checked `.golangci.yml`. Lint now reports FAILURE on this PR's two new lines — fixing them is should-fix-before-merge so the harness's own CI stays green and the regression suite keeps providing ongoing protection (independently caught by Jerry-Xin at the same head).

### Blocker
- `harness/gaptable/main.go:41` — `defer func() { _ = w.Close() }()` — repo's `.golangci.yml` has `errcheck.check-blank: true`, so `_ =` does NOT silence errcheck. Lint job logs: `Error return value of w.Close is not checked (errcheck)`. Fix: `defer func() { if err := w.Close(); err != nil { log.Printf("close kafka writer: %v", err) } }()`.
- `harness/gaptable/main.go:127` — `eb, _ := json.Marshal(encM)` — same rule, same outcome: `Error return value of json.Marshal is not checked (errcheck)`. Fix: `eb, err := json.Marshal(encM); if err != nil { log.Fatalf("marshal encrypted vector: %v", err) }`.

### Standing
All the praise + minor + out-of-scope notes from my earlier COMMENT (review 4545240208) still hold — harness design is solid (real assertions, real CDC path, locks #21 invariants, the `wantRawExcluded` 2→1 update is correct). Once the two errcheck lines flip Lint green, this should land.

yujiawei (commented):
# Code Review — PR #23 (octo-search-indexer)

**Verdict: CHANGES_REQUESTED** (correcting my earlier review at this same head `af9e308`). The harness logic is correct and well-designed, but the PR's own **Lint CI check is failing** on two real `errcheck` violations introduced by this PR. These must be fixed so the regression suite itself stays green and mergeable.

## Blocker — Lint FAILURE (errcheck, `check-blank: true`)

The repo's `.golangci.yml` sets `errcheck.check-blank: true`, so assigning an error return to `_` does **not** silence errcheck. `golangci-lint run ./harness/gaptable/...` reproduces exactly:

- `harness/gaptable/main.go:41` — `defer func() { _ = w.Close() }()` → *Error return value of `w.Close` is not checked (errcheck)*.
  Fix: `defer func() { if err := w.Close(); err != nil { log.Printf("close kafka writer: %v", err) } }()`
- `harness/gaptable/main.go:127` — `eb, _ := json.Marshal(encM)` → *Error return value of `json.Marshal` is not checked (errcheck)*.
  Fix: `eb, err := json.Marshal(encM); if err != nil { log.Fatalf("marshal encrypted vector: %v", err) }`

This was independently caught by two other reviewers at the same head. I am upgrading from my earlier verdict because my first pass ran `go vet` / `go build` (both green) but not `golangci-lint`, and so missed the Lint gate. Build / Test / Vet all pass; only Lint (and the unrelated `check-sprint` project-board gate) are red.

## Verification — harness logic (all ✅, unchanged from first pass)

- **`wantRawExcluded` 2→1 is correct**: image row `{"type":2,...}` (signal=0) → `docFromRow` (doc_enrich.go:35) → `ExtractVisibility` (no visibles key → broadcast) → `RawPayload` set → `DocFromMessage` branch A (doc.go:192) → `buildPayloadFromRaw` emits `payload.Image` (buildraw.go:81-98) → `projected=true` → `RawExcluded=false`. Only the Signal row stays `rawExcluded=true`. 6-row arithmetic holds (source=6, ES=5, rawExcluded=1, DLQ=1).
- **Visibility fail-closed (STEP 4)**: rows `[]` / `null` / `[123,456]` all error in `ExtractVisibility` → consumer pre-check (consumer.go:146) → DLQ `visibility_untrusted` (consumer.go:304), excluded from bulk; valid visibles indexed; no-key broadcast; encrypted row 020 → branch B indexed with empty visibles. Matches the contract.
- **Grepped strings real**: `mapping-compat FAILED` + `make migrate-forward` (mapping_compat.go:91-92), `es-indexer running` (cmd/es-indexer/main.go:67), `visibility_untrusted` (consumer.go:304). STEP0 fail-closed gate genuinely refuses startup via `os.Exit(1)`.
- **Snowflake precision / payloadRaw retention** assertions map to real `UseNumber()` + `rawForStore` behavior.

## Non-blocking (P2 / nit, from first pass — address opportunistically)

- **P2** Fixed `sleep 12`/`sleep 8` (gaptable-e2e.sh:117,128) → CI flakiness; prefer poll-until-ready.
- **P2** `KEEP_UP=1` skips `trap down EXIT` → leaked nohup indexer + uncleared consumer-group offsets.
- **P2** DLQ reason cross-check not pinned to expected `{11,12,13}` IDs and reads only partition 0 (safe only because topic is `--partitions 1`).
- **nit** Mapping check uses Python `assert` (gaptable-e2e.sh:107) — disabled under `PYTHONOPTIMIZE`; alias check is substring-only.

Once the two errcheck lines are fixed and Lint flips green, this is good to land — the harness adds real, well-targeted end-to-end coverage for the Plan B CDC-style write path and the P0 visibility fail-closed gate.

OctoBoooot (approved):
## Review: test(harness): gap-table end-to-end regression (#23) — delta review at e5da4655

**Verdict: Approve.**
Both errcheck blockers from af9e3081 are correctly fixed, plus the author proactively addressed the non-blocking minors my prior review (and Jerry-Xin's) raised about fixed-sleep flakiness and `set -euo pipefail` aborts. CI: check-sprint passes; Lint/Build/Test/Vet pending but the fix matches the existing `if err != nil` pattern (e.g. `rawMsg` at line 73) — high confidence it'll go green.

### Resolved since af9e3081 ✅
- `harness/gaptable/main.go:41-45` — `defer func() { _ = w.Close() }()` → `if cerr := w.Close(); cerr != nil { log.Printf("gaptable: writer close error: %v", cerr) }`. Correct shape for a non-fatal cleanup error.
- `harness/gaptable/main.go:131-134` — `eb, _ := json.Marshal(encM)` → `eb, err := json.Marshal(encM); if err != nil { log.Fatalf("marshal %s: %v", encM.MessageID, err) }`. Mirrors `rawMsg` at line 73.

### Bonus hardenings (above prior asks)
- `harness/gaptable-e2e.sh:62-79` — `jqid` now catches non-JSON / OpenSearch-error bodies and emits a `__SEARCH_ERR__` sentinel; `recall()` handles it as a fail-but-continue instead of letting `set -euo pipefail` abort the suite mid-run. Closes the prior risk Jerry-Xin raised.
- `harness/gaptable-e2e.sh:96-109` — Kafka readiness is now FAIL-LOUD (was best-effort `|| true`). Prevents the false-fail mode where a missing DLQ topic causes records to spill instead of being counted (the comment explains the failure chain explicitly — appreciated).
- `harness/gaptable-e2e.sh:142-152` — `sleep 12` replaced with a 60-tick poll on `es-indexer running` / `exited with error` / process death. Faster on fast hosts, more reliable on slow CI; surfaces startup failure immediately instead of waiting the full window.
- `harness/gaptable-e2e.sh:163-186` — `sleep 8` after seeding replaced with a poll waiting for BOTH "10 expected docs queryable" AND "DLQ delta == 3" to settle. The two-signal settle is the right design — neither alone catches a slow consumer that has indexed but not yet routed DLQ (or vice versa).
- `harness/gaptable-e2e.sh:213-220` — STEP 4 fail-closed loop now has `|| f=__SRC_ERR__` to keep curl/JSON failures from killing the suite; the FAIL message includes the actual `found` value for debuggability.

### Praise
- This is a model fix-cycle delta: addresses the named blockers, addresses the optional asks from both reviewers, *and* adds explanatory comments naming the failure mode each guard prevents (the Kafka-readiness FAIL-LOUD comment in particular explains exactly *why* `|| true` was unsafe). Future maintainers won't have to re-derive any of it.

Jerry-Xin (approved):
Summary: The PR is relevant to this repository and adds a useful live-path regression harness for the Kafka-to-OpenSearch CDC-style indexing contract; I found no blocking issues.

💬 Non-blocking:
- 🟡 Warning: `KEEP_UP=1` leaves the host `es-indexer` process running, not just the Docker stack. Because teardown is skipped at [harness/gaptable-e2e.sh](/Users/mlamp/.openclaw/tmp/pr-review-pdwlcbpw/repo/harness/gaptable-e2e.sh:111), `stop_indexer` is not called after the run. This can interfere with later local runs using the same fixed group, PID file, and topics. Consider always stopping the host indexer and only preserving Docker services/volumes.
- 🔵 Suggestion: The temp `BIN_DIR` created at [harness/gaptable-e2e.sh](/Users/mlamp/.openclaw/tmp/pr-review-pdwlcbpw/repo/harness/gaptable-e2e.sh:49) is never removed. Low impact, but a trap cleanup for script-created binaries would avoid `/tmp` buildup.

✅ Highlights:
- The relevance gate passes: this extends the project’s own indexing harness and backfill verification.
- Good coverage of strict mapping compatibility, `payloadRaw`, projected payload fields, snowflake precision, and visibility fail-closed behavior.
- The harness uses the real `cmd/es-indexer`, migration script, Kafka, and OpenSearch path rather than duplicating core behavior.
- `go test ./...` passed.

yujiawei (commented):
# Code Review — PR #23 (octo-search-indexer)

**Verdict: APPROVED** (no blocking findings). Re-reviewed against the **current head `e5da465`** (merge-base `main`). This is a re-review: the prior round requested changes at `af9e308` for two `errcheck` lint violations; this head fixes both and additionally hardens the harness. Scope is unchanged — a test harness only (gap-table e2e shell driver + Kafka seeder + one constant in the backfill harness). **No production code is modified.**

## 1. Verification

- ✅ **All CI green at this head.** Build / Test / Vet / Lint / CodeQL / osv-scanner / secret-scan / dependency-review all pass. The previously-red `Lint` and the `check-sprint` project-board gate are now green.
- ✅ **Both prior `errcheck` blockers are genuinely fixed** (repo `.golangci.yml` has `errcheck.check-blank: true`, so `_ =` does *not* silence it):
  - `harness/gaptable/main.go:41` — `w.Close()` is now checked: `defer func(){ if cerr := w.Close(); cerr != nil { log.Printf(...) } }()`. Non-fatal `Printf` is the correct choice for deferred cleanup.
  - `harness/gaptable/main.go:128` — `json.Marshal(encM)` is now checked with `log.Fatalf` on error. Fatal is correct for an unrecoverable seed-prep failure, and the message carries the `MessageID` for context.
  - Local `go build ./harness/...`, `go vet ./harness/...`, and `bash -n harness/gaptable-e2e.sh` all clean.
- ✅ **Assertion semantics preserved — the delta changed only timing, error-handling, and infra setup, never a check.** Independently confirmed:
  - No `recall`/`src` target id or expected value changed.
  - The fixed `sleep 12` / `sleep 8` were replaced by bounded poll loops, but the readiness gates (`present -ge 10`, `dlq_delta -ge 3`) are *gates only*; the **exact** DLQ assertion `DLQ_NEW -eq 3 && DLQ_VU -eq 3` (`gaptable-e2e.sh:229`) still stands.
  - Kafka readiness and topic creation moved from best-effort (`|| true`) to **fatal** (`exit 1`). This is a correctness improvement: a missing `octo.message.v1.dlq` topic would otherwise let poison records spill to disk and silently false-fail the STEP4 fail-closed count. Failing loud is the right call.
  - The new `recall`/`src` error paths (`__SEARCH_ERR__` / `__SRC_ERR__` sentinels) correctly convert a non-search response into a counted `FAIL` instead of aborting the suite under `set -euo pipefail` — strictly safer than before.
- ✅ **Underlying harness logic (carried from the prior verified pass at `af9e308`, unchanged here):** `wantRawExcluded` 2→1 is correct (image row projected → `RawExcluded=false`; only the Signal-encrypted row stays raw-excluded); visibility fail-closed routing (`[]` / `null` / non-string `visibles` → DLQ `visibility_untrusted`, not fail-OPEN); snowflake `>2^53` precision via `UseNumber()`; `payloadRaw` retention. All assertions map to real production behavior.

## 2. Findings (all non-blocking)

- **nit — `INDEXER_PID="$(cat "$INDEXER_PID_FILE")"` (`gaptable-e2e.sh:145`) is the one new command-substitution not given a `|| fallback` guard.** The sibling new assignments (`dlq_now`, `present`, `f`) all guard against `set -euo pipefail` abort, but this one does not. In practice it is safe: line 141 (`echo $! > "$INDEXER_PID_FILE"`, fixed `/tmp` path) writes the file immediately before, so `cat` cannot realistically fail; and if `/tmp` were unwritable the script would already abort at line 141 — i.e. before any assertion, at the infra-precondition layer. For symmetry with the other guarded reads, consider `INDEXER_PID="$(cat "$INDEXER_PID_FILE")" || INDEXER_PID=""`. Not a blocker.

## 3. Suggestions (carry-overs, opportunistic)

- The DLQ reason cross-check counts exactly 3 records but does not pin *which* three IDs (`{11,12,13}`) landed; the independent per-ID absence checks already cover this, so it's a residual-margin tightening only.
- `KEEP_UP=1` skips `trap down EXIT`, leaving the `nohup` es-indexer running; consider always stopping the indexer on exit even when keeping the stack.

## 4. Security note (this PR was auto-classified `security_sensitive`)

The classification is keyword-triggered (the test vectors exercise `visibles` / encrypted-DM / fail-closed paths). The actual change is **test-harness only — zero production-code surface**, connecting to a throwaway local stack with fixed local ports, never wired into any shared environment. The security-relevant behavior it *exercises* (visibility fail-closed → DLQ) lives in already-merged production code (#21) and is unchanged here. Nothing in this diff weakens a safety gate; if anything the fatal-on-missing-DLQ-topic change strengthens the test's ability to catch a fail-OPEN regression. No manual security action required for this diff.

This is a clean, well-scoped re-review pass: the requested lint fixes landed correctly, the robustness hardening only strengthens the suite, and CI is fully green. Good to land.

mochashanyao (approved):
[Octo-Q · automated review]

---

# octo-search-indexer PR#23 Review Report
**PR:** https://github.com/Mininglamp-OSS/octo-search-indexer/pull/23
**Head SHA:** `e5da4655f2c88cca13a07c9b05947f7de1b8c876`
**Title:** test(harness): gap-table end-to-end regression (CDC-style writes)
**Scope:** 3 files, +398/-4 (harness test code only, zero production changes)

---

## 1. Verification Summary

| Item | Status | Evidence |
|------|--------|----------|
| backfill wantRawExcluded 2→1 | ✅ | `doc.go:203` Branch A: `RawExcluded = !projected`; image type=2 → `buildPayloadFromRaw` produces `payload.image` → `projected=true` → `RawExcluded=false`. Only Signal-encrypted row remains raw_excluded. |
| gaptable seeder message contract | ✅ | `gaptable/main.go` uses `searchmsg.SchemaVersion` (==2), `SourceETLMessageTable`, `json.RawMessage` for RawPayload — matches consumer `decodeAndValidate` strict equality check. |
| Encrypted DM vector (id …020) | ✅ | `RawExcluded:true, Content:nil, RawPayload:nil` → Branch B (`doc.go:204-207`): `Payload=nil, RawExcluded=true`. Test asserts `rawExcluded && !payload && !visibles`. |
| Visibility fail-closed vectors (V1-V5) | ✅ | Covers all 4 `ExtractVisibility` return cases: valid array→indexed, empty array→DLQ, null→DLQ, all-non-string→DLQ, absent key→broadcast allowed. |
| STEP0 stale-mapping gate | ✅ | `AssertLiveMappingCompatible` (`mapping_compat.go:47`) checks required paths + `payloadRaw` enabled:false. Stale mapping lacks these → error → indexer exits non-zero. Test checks both exit code and log message. |
| DLQ fencing | ✅ | Captures `DLQ_START` offset before seeding, counts only records at/after it. Prevents stale DLQ records from prior runs causing false failures. |
| Mapping fields (spaceId, visibles) | ✅ | `octo-message.json:18-19` confirms both fields exist as `keyword` type in v1.9 mapping. |
| Snowflake precision (9007199254740993) | ✅ | `buildPayloadFromRaw` uses `decodeObjectUseNumber()` (`buildraw.go:180`) preserving >2^53 precision. Python `json.load` also handles arbitrary-precision ints. |

## 2. Findings

### No P0 or P1 findings.

### P2 — Minor observations (non-blocking)

**P2-1: `base` parameter unused in `vectors()`**
- `gaptable/main.go:89`: `base := time.Now().Unix()` is passed to `vectors(base)` but the `base int64` parameter is never referenced in the function body. All timestamps are hardcoded constants (1700000123, 1700000456) or not set.
- **Diff-scope:** new (introduced in this PR).
- **Impact:** Cosmetic — no functional effect. The seeder works correctly without it.
- **Suggestion:** Either remove the parameter or use it for the `MsgTimestamp`/`CreatedAt` fields to make the test data time-relative.

**P2-2: `docs_present` makes 10 sequential HTTP calls per poll tick**
- `gaptable-e2e.sh` lines ~165-172: Each poll iteration makes one `curl` + `python3` invocation per expected doc (10 total), up to 40 iterations.
- **Diff-scope:** new.
- **Impact:** Performance — worst case ~400 subprocess spawns. Acceptable for a throwaway harness; not a concern for correctness.
- **Suggestion:** Could use `_mget` API for batch doc fetch if latency becomes an issue in CI.

**P2-3: `|| true` on DLQ consumer could mask a real Kafka error**
- `gaptable-e2e.sh` line ~220: `grep -c 'visibility_untrusted' || true` swallows grep failures (e.g., consumer timeout with 0 messages).
- **Diff-scope:** new.
- **Impact:** Low — the subsequent `[ "$DLQ_NEW" -eq 3 ] && [ "$DLQ_VU" -eq 3 ]` assertion would still fail if the consumer silently errored. The `|| true` only prevents `set -e` abort, not the assertion.

## 3. Suggestions

- Consider removing the unused `base` parameter from `vectors()` or using it for timestamps (P2-1).
- The test is well-structured with proper polling-based settlement detection, DLQ offset fencing, and comprehensive coverage of the `ExtractVisibility` fail-closed contract. No structural changes needed.

## 4. Additional Observations

- **Test isolation is solid:** Throwaway Docker stack, dedicated consumer group (`octo-gaptable-e2e`), DLQ offset fencing, and `trap down EXIT` cleanup.
- **26 assertions verified:** Counted across STEP0(1) + STEP1(2) + STEP2(1) + STEP3(16) + STEP4(6) = 26, matching the PR body claim.
- **The backfill fix is correctly scoped:** Only changes the assertion constant and comments/log message. No logic changes. The root cause (image rows now projected under CDC-style writes) is well-documented in the updated comments.

## 5. Data Flow Traceback

|---|---|---|
| `wantRawExcluded=1` | `doc.go:203` `RawExcluded = !projected` where `projected = hasTypedSubObject()` from `buildPayloadFromRaw` | ✅ Image type=2 produces `payload.image` → projected=true → RawExcluded=false |
| Gap-table vectors (G1-G7) | `gaptable/main.go` `rawMsg()` → Kafka → consumer `processBatch` → `ExtractVisibility` + `buildPayloadFromRaw` → ES bulk | ✅ All fields flow through projection to indexed doc |
| V1 valid visibles | `ExtractVisibility` case 4: valid string array → `(spaceID, visibles, nil)` | ✅ Indexed with `visibles=['u_admin1','u_admin2']` |
| V2/V3/V4 poison visibles | `ExtractVisibility` cases 2/3: empty/null/non-string → `err != nil` (fail-closed) | ✅ Consumer sets `visibilityInvalid[i]` → DLQ reason=`visibility_untrusted` |
| V5 broadcast | `ExtractVisibility` case 1: no `visibles` key → `(spaceID, nil, nil)` | ✅ Allowed through, no gate triggered |
| Encrypted DM (id …020) | `RawPayload:nil, RawExcluded:true` → Branch B (`doc.go:204`) | ✅ `Payload=nil, RawExcluded=true`, skips visibility pre-check (len(RawPayload)==0) |
| `spaceId` in G7 | Payload JSON `"space_id":"space_top"` → `ExtractVisibility` extracts spaceID → backfills `parsed[i].SpaceID` | ✅ Mapping has `spaceId` keyword field |
| Snowflake 9007199254740993 | `buildPayloadFromRaw` uses `UseNumber()` decoder → preserves precision through JSON round-trip | ✅ Python `json.load` handles arbitrary-precision int |

- **C1 Dual-path parity:** N/A — test-only code, no symmetric add/remove paths.
- **C2 Control-flow ordering / nested reuse:** N/A — test harness functions are not reused across multiple call-sites with different orderings.
- **C3 Authorization boundary ≠ capability boundary:** N/A — harness code connects only to local throwaway Kafka/OpenSearch. No authorization boundaries.
- **C4 Authorization lifecycle / container-member state cascade:** N/A — no auth changes.
- **C5 Build/note ≠ runtime path:** N/A — test harness, not a build artifact. The `go build` step produces test binaries that run against the local Docker stack. No packaging/deployment path involved.
- **C6 Governance/policy/security doc self-consistency:** N/A — no documentation changes.

## 7. Cross-round Blocker Recheck (R6)

---

**[Octo-Q] verdict: APPROVE**

This is a well-engineered test harness with zero production code changes. The backfill assertion fix (wantRawExcluded 2→1) is correctly justified by the CDC-style projection behavior change. The gap-table E2E test provides comprehensive coverage of the Plan B live path including the critical visibility fail-closed security gate. All 26 assertions are sound, DLQ fencing prevents cross-run interference, and the stale-mapping gate test validates the migrate-forward prerequisite. Only minor P2 observations (unused parameter, sequential HTTP calls, defensive `|| true`).

PR #39, opened 2026-06-29
feat(obs): add Prometheus metrics for producer and consumer

### LINKED ISSUE #38: [observability] producer 指标缺 Histogram、consumer 完全无 metrics/obs
## 背景

当前可观测性两条腿不对等，运维侧拿不到关键延迟分布：

### Producer (`searchetl-producer`)
- `internal/producer/metrics.go` 是手写 `atomic` + `strings.Builder` 拼 Prometheus 文本，**无 client SDK、无 Histogram**
- 只有计数，拿不到 tick / read-batch 的耗时分布，无法做 p99 延迟告警
- 表里规划过的 `produce_errors_total` / `lock_renew_failures_total` / `dlq_total{reason}` 实际未埋点

### Consumer (es-indexer)
- `internal/consumer/` **完全没有 metrics，也没有 obs server / `/metrics`**
- 只有 `service.go` 里占位的 `logAlerter`（打 `[ALERT]` 日志），无任何 Prometheus 序列
- 索引落库延迟、bulk 失败、DLQ、offset lag 都不可观测

## 建议优化

1. 引入 `prometheus/client_golang`，每条腿独立 `Registry`（避免两 binary 互相误注册指标）
2. **Producer**：metrics.go 迁到 SDK，补 `tick_duration` / `read_batch_duration` 两个 Histogram，补齐 errors / lock / dlq 计数
3. **Consumer**：从零建 obs server（`/healthz` `/readyz` `/metrics`）+ 7 个 `indexer_*` 序列，`logAlerter` -> `metricsAlerter`（接口不变）
4. 约束：**纯埋点 + obs 基建，不碰正确性路径**（事务边界 / cursor CAS / DLQ 决策 / offset 提交一律不动）

## 备注

fork (`mou-ren`) 已有完整实现（`feat/observability-metrics`），含 producer 8 / consumer 7 指标 + 单测，可整理成 PR 提上来。

mou-ren:
Closes #38

## 做了什么

为 producer / consumer 两条腿补齐 Prometheus 可观测性，纯埋点 + obs 基建，**不碰正确性路径**（事务边界 / cursor CAS / DLQ 决策 / offset 提交一律未改）。

### Producer (`searchetl-producer`)
- `internal/producer/metrics.go` 由手写 `atomic` + 文本拼接迁到 `prometheus/client_golang`（私有 Registry）
- 删 `last_tick_unixtime`；`ticks` 两计数合并为 `ticks_total{result}`
- 新增 `tick_duration` / `read_batch_duration` 两个 Histogram
- 补齐 `produce_errors_total` / `lock_renew_failures_total` / `dlq_total{reason}`

### Consumer (es-indexer)
- 从零建 metrics.go + obs.go（原先完全无 metrics / obs server）
- `logAlerter` -> `metricsAlerter`（接口不变），7 个 `indexer_*` 序列
- `cmd/es-indexer` main 暴露 obs（`/healthz` `/readyz` `/metrics`）

### 通用约定
- 每条腿独立 `prometheus.NewRegistry()`，避免两 binary 互相误注册
- Histogram buckets 统一 `[.005 .01 .025 .05 .1 .25 .5 1 2.5 5]`
- label 基数克制：`result/op/reason/disp/partition/stream/shard`

## 自测
- `go build ./...` 通过
- `go test ./internal/producer/... ./internal/consumer/...` 全绿（含新增 metrics 单测）
- 已 rebase 到最新 `upstream/main`

落地方案细节见 `docs/observability-metrics-plan.md`。


OctoBoooot (commented):
## Review: feat(obs): add Prometheus metrics for producer and consumer (#39)

**Verdict:** Comment (CI red — can't approve until green)

CI is red (2 checks failing: `Lint`, `check-sprint`) — that needs fixing first. The code itself is a clean, well-scoped observability change, so this is a COMMENT, not a REQUEST_CHANGES: fix the lint and it's in good shape. Risk tier: high (queue consumers / Kafka) — but the change is purely additive instrumentation; I verified it doesn't touch the correctness path (tx boundaries / cursor CAS / DLQ decisions / offset commit all unchanged), matching the PR's claim.

### Blocker (CI)

- `internal/consumer/metrics_test.go:139` — `errcheck` lint failure: `_ = p.processBatch(...)` trips the linter because `.golangci.yml:7` sets `errcheck: check-blank: true` (assigning an error return to `_` is itself a violation). The sibling call at `:106` does it right (`if err := p.processBatch(...); err != nil { t.Fatalf }`). Mirror that here — the test should assert the call succeeds anyway. Until Lint goes green the verdict can't be APPROVE.

### Notes (non-blocking, all good)

- **Label cardinality is disciplined** — every `*Vec` label is a bounded enum (`disp`/`reason`/`op`/`result`/`stream`/`shard`) or Kafka-partition-bounded (`partition`/`committed_offset`). No unbounded-label series (the classic Prometheus memory-blowup), and the PR body's "label 基数克制" claim checks out.
- **Private registries per binary** (`prometheus.NewRegistry()` in each leg, no global default) correctly prevents the two binaries cross-registering — good call, and it's also what makes the per-test `NewMetrics()` isolation work.
- **Obs server is safe** — `ReadHeaderTimeout: 5s` (slow-loris guard), `/metrics` served from the private registry via `promhttp.HandlerFor`, graceful `Shutdown`.
- **nil-guarded hooks** (`if p.metrics != nil`) keep the instrumentation from coupling the processor to metrics being wired — clean additive seam.

### Praise

- The metric hooks are genuinely additive: each `markDisposition`/`ObserveIO`/`MarkIOError` sits *beside* the existing disposition/bulk logic without altering a single control-flow decision, and both legs got real `metrics_test.go` coverage asserting the emitted series. For a "pure instrumentation" PR on a high-risk consumer, keeping the correctness path provably untouched is exactly right.

(check-sprint is the governance/sprint-assignment gate, orthogonal to the code — but it's still a required check, so it also needs resolving before merge.)

mochashanyao (dismissed):
[Octo-Q · automated review]

---

# octo-search-indexer PR#39 Review Report
**PR:** feat(obs): add Prometheus metrics for producer and consumer
**Head SHA:** `23af214f75ca00f5a1c204339cfde8728542910b`
**Files changed:** 21 (+1081 / -122)
**Build:** ✅ `go build ./...` clean
**Tests:** ✅ `go test ./...` all pass (consumer/producer/esindex/backfill/recon)
**Vet:** ✅ `go vet ./...` clean

---

## 1. Verification Summary

| Item | Status | Evidence |
|---|---|---|
| Build compiles | ✅ | `go build ./...` exit 0 |
| All tests pass | ✅ | `go test ./...` — all packages OK |
| go vet clean | ✅ | no diagnostics |
| Producer metrics SDK migration | ✅ | `internal/producer/metrics.go` — atomic→CounterVec/GaugeVec/Histogram, private registry, promhttp handler |
| Consumer metrics from scratch | ✅ | `internal/consumer/metrics.go` + `obs.go` — 8 series, private registry, nil-safe injection |
| Producer obs server idle path | ✅ | `cmd/searchetl-producer/main.go serveIdleObs` — /healthz + /readyz 200, graceful shutdown |
| Consumer obs server idle path | ✅ | `cmd/es-indexer/main.go serveIdleObs` — mirrors producer pattern |
| Consumer obs server enabled path | ✅ | `cmd/es-indexer/main.go` L75-87 — NewObsServer + Start + defer Shutdown |
| Metrics injection chain (producer) | ✅ | main→NewMetrics→ETL.d.Metrics→runLocked→runChunk; main→NewKafkaProducerWithMetrics(cfg, metrics) |
| Metrics injection chain (consumer) | ✅ | NewService→NewMetrics→newKafkaSource(metrics)→newKafkaDLQSink(metrics)→NewProcessor(metrics) |
| Nil-safety on all metrics call sites | ✅ | every `metrics.X()` is guarded by `if metrics != nil` |
| DLQ counting (producer) | ✅ | `etl.go` L190: MarkDLQ only after ProduceDLQ success — no double-count on re-produce |
| DLQ counting (consumer) | ✅ | `consumer.go` routePoison: MarkDLQ only after dlq.Send success; hard_stop via alerter only |
| alerter event→counter mapping | ✅ | `service.go` metricsAlerter: bulk_batch_error→MarkBulkError, dlq_hard_stop→MarkDLQHardStop, dlq_write_exhausted→MarkDLQWriteExhausted — semantically separated, no double-count |
| Lock renew failure counting | ✅ | `lock.go` renewUntilDone: MarkLockRenewFailure on both err and ownership-lost paths |
| kafka_fetch IO error filtering | ✅ | `kafka.go` L60: ctx cancellation excluded from IO error count (correct — normal shutdown) |
| Test coverage (metrics unit) | ✅ | `producer/metrics_test.go` + `consumer/metrics_test.go` — dumpMetrics helper, counter/gauge/histogram assertions |
| Test coverage (integration) | ✅ | `consumer/metrics_test.go TestProcessBatch_DispositionMetrics` — real processBatch pipeline, not hand-fed |
| Test coverage (semantic lock) | ✅ | `TestDLQHardStopVsExhausted_MetricSemantics` — verifies hard_stop vs exhausted distinction across spill/no-spill |

---

## 2. Findings

### F1 — P2: Consumer `disposition_total` over-counts on transient retry rounds

**Diff-scope:** new (introduced by this PR's instrumentation)

**Location:** `internal/consumer/consumer.go` L273 (in resolvePass bulk-result loop):
```go
default:
    dispositions[idx] = dispTransient
    p.markDisposition(dispTransient)  // counts each transient item per pass
```

**Trace:** `processBatch` runs `resolvePass` in a retry loop (L186: `for attempt := 0; ; attempt++`). Each pass calls `markDisposition` for every item with a new outcome. An item that is transient on pass 1 then OK on pass 2 produces two metric increments: `disposition_total{transient}+1` then `disposition_total{ok}+1`. For a batch of N items where K are transient-then-resolved, total disposition increments = N + K.

**Impact:** Metric inflation, not correctness. `rate(indexer_disposition_total{disp="ok"})` will exceed the actual item throughput when transient retries are common. Alerting on `ok / total` ratio would give misleading health signals under load.

**Severity:** P2 (metric inaccuracy, no data loss, no user-visible error, no correctness path affected).

**Suggestion:** Either (a) only count terminal dispositions (skip `markDisposition(dispTransient)` in the bulk-result default case), or (b) rename the metric doc to clarify it's per-pass not per-item.

---

### F2 — P2: Consumer obs `SetReady(true)` before startup health checks

**Location:** `cmd/es-indexer/main.go` L82-83:
```go
obs.SetReady(true)  // set before svc.Run()
// ...
return svc.Run(ctx)  // EnsureIndex + AssertLiveMappingCompatible inside
```

**Trace:** `svc.Run()` at `service.go` L140 calls `EnsureIndex` and `AssertLiveMappingCompatible` before entering the consumer loop. If either fails, `Run` returns an error and the process exits. But `/readyz` already returns 200 during that window (typically sub-second).

**Impact:** Minimal — the window is tiny, no HTTP requests are in flight, and the pod exits on failure. Kubernetes would remove it on the next failed probe. Not a user-visible issue.

**Severity:** P2 (technically incorrect readiness signal, negligible practical impact).

**Suggestion:** Move `obs.SetReady(true)` to after `svc.Run()` completes its startup checks (would require either a post-init callback or restructuring the run flow). Low priority.

---

## 3. Suggestions

1. **F1 fix:** Consider removing `p.markDisposition(dispTransient)` from the bulk-result default case in `resolvePass`. Only terminal dispositions (ok/dlq) need counting; transient retries are implicit from `io_op_errors_total{es_bulk}`.
2. **Removed metric note:** `searchetl_producer_last_tick_unixtime` was removed without replacement. If any dashboard/alert depends on it, it will silently stop updating. Confirm no downstream consumers rely on it.
3. **Minor:** The `docs/observability-metrics-plan.md` is a thorough design doc — good to ship alongside the code for future maintainers.

---

## 4. Additional Findings

- **No cross-package registry collision:** Producer uses namespace `searchetl_producer`, consumer uses `indexer`. Both use private `prometheus.NewRegistry()` — even if accidentally loaded in the same binary, no `MustRegister` panic.
- **Label cardinality is restrained:** All label dimensions have ≤4 values (disp=3, op=4, result=2, reason≤3, stream=2). No unbounded labels.
- **Idle obs server symmetry:** Both producer and consumer idle paths serve /healthz + /readyz with 200 and nil metrics — consistent, prevents crashloop for deliberately-disabled pods.

---

## 5. Data-Flow Tracing

| Consumed Data | Upstream Source | Flow Verified |
|---|---|---|
| `metrics` in `runChunk` | `e.metrics` from ETL struct → ETLDeps.Metrics from `main.go` `producer.NewMetrics()` | ✅ Non-nil in production, nil in tests |
| `metrics` in `renewUntilDone` | `runLocked` param → `ETL.RunIncremental` passes `e.metrics` | ✅ Nil-safe guard at each call site |
| `metrics` in `KafkaProducer` | `NewKafkaProducerWithMetrics(cfg, metrics)` in `main.go` sink factory | ✅ Flows from same `producer.NewMetrics()` instance |
| `svc.Metrics()` in es-indexer `main.go` | `Service.metrics` ← `NewService` → `NewMetrics()` | ✅ Same instance shared with Processor, kafkaSource, kafkaDLQSink |
| `MarkDLQ(plan.dlq[i].Reason)` | `DLQEnvelope.Reason` set by `planChunk` → `extractMessage` outcome | ✅ String constants: `producer_payload_unparseable` / `producer_visibility_untrusted` / `producer_oversize_truncated` |
| `MarkDLQ(reason.reasonString())` (consumer) | `dlqReason` enum → `reasonString()` method | ✅ Mapped: `unknown_schema_version` / `visibility_untrusted` / `permanent_4xx` |
| `metricsAlerter.Alert("bulk_batch_error", ...)` | `resolvePass` L245 → `alert.Alert` → `MarkBulkError()` | ✅ Single-count per batch-level bulk failure |
| `metricsAlerter.Alert("dlq_hard_stop", ...)` | `routePoison` L295 → `alert.Alert` → `MarkDLQHardStop()` | ✅ Only on DLQ terminal escape (no spill), not double-counted with MarkDLQ |
| `metricsAlerter.Alert("dlq_write_exhausted", ...)` | `dlqHandler.Send` L112 → `alert.Alert` → `MarkDLQWriteExhausted()` | ✅ Fires once per retry-exhaustion, independent of hard_stop vs spill outcome |
| `ObserveIO("kafka_fetch", ...)` | `kafkaSource.Fetch` L57 → `reader.FetchMessage` timing | ✅ ctx cancellation excluded from error count |
| `SetCommittedOffset(partition, offset)` | `commitPrefixes` L309 → after `src.Commit` success | ✅ Only on successful commit, partition as string label |

---

## 6. Blind-Spot Checklist (R5)

- **C1 Dual-path parity:** N/A — No symmetric add/remove or create/delete paths touched. Metrics instrumentation is additive, no behavioral symmetry to verify.
- **C2 Control-flow ordering / nesting reuse:** CLEAR — `ObserveReadBatch` wraps only `ReadStableBatchTx` (single IO call, no nesting). `ObserveTickDuration` wraps `tickFn` (single invocation per tick). No double-observation risk. `MarkTick` called once per tick with the correct result label.
- **C3 Authorization boundary ≠ capability:** N/A — No auth/permission/jail/tool changes. Pure observability addition.
- **C4 Authorization lifecycle / container-member cascade:** N/A — No auth changes.
- **C5 Build ≠ runtime path:** N/A — No build artifacts, extensions, CLI packaging, or relative URL changes. Two separate binaries (`searchetl-producer`, `es-indexer`) each with their own private registry — verified no cross-registration.
- **C6 Governance / policy doc consistency:** N/A — The added `docs/observability-metrics-plan.md` is a technical design doc, not a governance/policy document. No security or disclosure policy changes.

---

## 7. Cross-Round Blocker Review (R6)

---

[Octo-Q] verdict: APPROVE

This PR cleanly migrates producer metrics from hand-rolled atomics to the Prometheus client_golang SDK, and adds consumer metrics + obs server from scratch. All metrics injection paths are nil-safe, DLQ counting is correctly post-confirmation (no double-count on re-produce), alerter event→counter mapping is semantically precise (hard_stop vs exhausted correctly separated), and test coverage includes both unit tests and integration tests through real code paths. No P0/P1 findings. Two P2 metric-accuracy observations noted (transient over-count on retry, readiness timing) that don't affect any correctness path.

Jerry-Xin (dismissed):
Relevance gate: pass. This PR is in scope for `octo-search-indexer`; it adds producer/consumer observability for the existing indexing pipeline. No blocking issues found.

**💬 Non-blocking**
- 🟡 Warning: `cmd/es-indexer/main.go:74-87` marks `/readyz` ready before `svc.Run` performs `EnsureIndex` and mapping compatibility checks. During a slow or failing startup, readiness can return 200 before the consumer is actually ready. Consider flipping ready only after startup gates pass.
- 🟡 Warning: `internal/consumer/consumer.go:296-299` records `dlq_hard_stop_total` for any `dlq.Send` error. `dlq.Send` can return `context.Canceled`/deadline errors from shutdown paths, not only `errDLQHardStop`, so this can create false hard-stop alerts. Prefer `errors.Is(err, errDLQHardStop)`.
- 🔵 Suggestion: `internal/consumer/consumer.go:270-272` increments `disposition_total{disp="transient"}` on every retry pass. A message that eventually succeeds can be counted as both transient and ok. If that is intentional, document it as attempt-level disposition; otherwise count only terminal outcomes or state transitions.

**✅ Highlights**
- Producer metrics migration keeps a private Prometheus registry and avoids global collector pollution.
- Consumer metrics are wired through the real processing path with focused tests.
- `go test ./...` passes.

OctoBoooot (commented):
## Follow-up (same head `23af214f`) — confirming a peer metrics-accuracy finding

My prior review (4588847687) stands: **Comment**, CI red (Lint errcheck + check-sprint) must go green before merge. This adds one finding a peer raised that I'd missed — same head, no verdict change.

### Minor (new) — `disposition_total{disp="transient"}` counts per retry-attempt, not per item

`internal/consumer/consumer.go` — verified: `processBatch` reruns `resolvePass` on the same batch each retry attempt (`:164` loop, in-place retry, offset not advanced) until no item is transient. In `resolvePass`, the terminal dispositions are correctly guarded — `dispOK`/`dispDLQResolved` fire only on the transient→terminal transition, after which the item is skipped on later passes, so **ok/dlq do NOT double-count**. But the `default` branch (`:271`) calls `markDisposition(dispTransient)` for every still-transient item on **every** pass. So a single item that takes N attempts to resolve emits N `transient` increments while contributing exactly one `ok`/`dlq` — the three labels of `disposition_total` end up with different denominators (they no longer sum to item count).

Impact: observability-only (no correctness/offset/DLQ effect), which is why this is a **minor**, not a major. But under a counter named `disposition_total` it's misleading — a dashboard computing `ok / sum(disposition_total)` would under-report success rate under retry pressure.

Direction (author's call): either (a) only `markDisposition(dispTransient)` on the *first* pass (mark transient once when the item enters the batch, then only emit terminal dispositions thereafter), or (b) keep the per-attempt count but move it to a distinctly-named series like `transient_retries_total` so it's not conflated with terminal per-item dispositions. Add a retry-loop test asserting a 2-attempt item yields `transient`≤1 (or the intended count) + exactly one terminal.

Credit to mochashanyao for spotting the inflation — I'd checked the terminal-disposition guards and the label cardinality on the first pass but didn't trace the transient mark through the retry loop. Lesson: when a counter is incremented inside a retry/until-converged loop, check each label's increment is once-per-entity, not once-per-iteration.

Jerry-Xin (changes requested):
CI is red on this head, so this can't be approved yet. The metrics design itself is sound (see highlights), but there's a real lint blocker plus a non-code gate.

**🔴 Blocking**
- 🔴 CI Lint (errcheck) is failing: `internal/consumer/metrics_test.go:139:3` — return value of `p.processBatch` is not checked (`errcheck`). This is introduced by the new test code in this PR. Fix it (e.g. `if err := p.processBatch(...); err != nil { t.Fatalf(...) }`, or assign to `_` only if intentional and allowed by the linter config). The Lint check must be green before merge.

**💬 Non-blocking**
- 🟡 `cmd/es-indexer/main.go:74-87` marks `/readyz` ready before `svc.Run` performs `EnsureIndex` and mapping compatibility checks. During a slow or failing startup, readiness can return 200 before the consumer is actually ready. Consider flipping ready only after startup gates pass.
- 🟡 `internal/consumer/consumer.go:296-299` records `dlq_hard_stop_total` for any `dlq.Send` error. `dlq.Send` can return `context.Canceled`/deadline errors from shutdown paths, not only `errDLQHardStop`, so this can create false hard-stop alerts. Prefer `errors.Is(err, errDLQHardStop)`.
- 🔵 `internal/consumer/consumer.go:270-272` increments `disposition_total{disp="transient"}` on every retry pass. A message that eventually succeeds can be counted as both transient and ok. If intentional, document it as attempt-level disposition; otherwise count only terminal outcomes/state transitions.
- ℹ️ `check-sprint` is also failing, but that's a non-code governance gate (the linked issue has no Sprint set on the Octo Board) — needs a Maintainer to assign it to the current sprint, not a code change.

**✅ Highlights**
- Cardinality is well-controlled: every metric label value traces to a bounded enum/const (stream, result, reason, shard=ETL table name, disp, partition=Kafka partition int, op) — no user_id/doc_id/error-string/raw-query labels, so no cardinality blowup.
- Each leg uses its own private `prometheus.NewRegistry()` + single `MustRegister` in `NewMetrics()` called once per service — no global collector pollution, no per-request/double-register panic.
- Histograms observe `d.Seconds()` against 5ms–5s buckets (correct unit); single emission path per counter (routePoison explicitly does not re-Inc); `kafka_fetch` IO-error guarded by `ctx.Err()==nil` to avoid counting graceful shutdown.
- `/metrics` exposed on a separate internal obs port (`INDEXER_OBS_ADDR`), standard sidecar-scrape pattern.
- Real test coverage: `Registry().Gather()` + exact series-string assertions, plus a semantic regression test for hard-stop vs exhausted.

yujiawei (dismissed):
# Code Review — PR #39 (octo-search-indexer)

**Head SHA reviewed:** `23af214f75ca00f5a1c204339cfde8728542910b`
**Base / merge-base:** `7fedf7de298649e28372038bf2e57c21a2181c03`

Reviewed the full diff against the checked-out tree, traced every metric to its firing site, and ran the toolchain locally at this head SHA: **`go build ./...`, `go vet ./...`, and `go test ./...` all pass** (race + shuffle covered by CI). Verification was done across four independent dimensions (correctness, cardinality, concurrency, scope) with adversarial re-checking of each finding.

## Verdict

**APPROVED.** No correctness, data-loss, security, or build-blocking issues. The change is cleanly scoped to instrumentation + obs plumbing and does not touch the load-bearing correctness paths (ETL transaction boundaries, cursor CAS, offset commit, DLQ routing decisions). The findings below are observability-accuracy refinements only and do not block merge.

---

## 1. Scope / Spec compliance — PASS

- **Producer:** exactly the 8 planned series registered (`produced_total{stream}`, `cursor_position{shard}`, `ticks_total{result}`, `tick_duration_seconds`, `read_batch_duration_seconds`, `dlq_total{reason}`, `produce_errors_total`, `lock_renew_failures_total`). The planned deletions are gone repo-wide (`last_tick_*`, `tick_errors_total`, hand-written `Render()`), and the `searchetl_producer_` prefix is preserved across the SDK migration so existing dashboards keep working. Every series is wired to a live code path — no dead metrics.
- **Consumer:** the 7 planned series are all present and wired, plus **one extra** (`dlq_write_exhausted_total`). This +1 is **justified, not scope creep**: the original plan mapped both the `dlq_hard_stop` and `dlq_write_exhausted` alert events onto a single counter, which would double-count on a real no-spill hard stop (exhausted fires, then `escape()` returns `errDLQHardStop` → hard_stop fires too) and false-positive when spill is configured. Splitting them into two counters is a correctness improvement over the plan's own mapping, wired to a pre-existing alert event (`dlq.go` is unchanged) and locked by `TestDLQHardStopVsExhausted_MetricSemantics`.
- **Non-intrusion verified:** `runChunk` / `commitPrefixes` / `resolvePass` only gain a `metrics` param + timing wrappers around existing calls; SQL/FOR UPDATE/Commit/keyset/CAS and the commit/DLQ branches are untouched. `logAlerter → metricsAlerter` keeps the `Alert(event, detail)` interface. `NewKafkaProducer` is retained as a nil-metrics shim, so the change is backward-compatible. `INDEXER_OBS_ADDR` and the idle-obs server mirror the existing producer leg as the plan specifies — no unrequested flags.
- **Cardinality / registry hygiene — clean:** both binaries use private `prometheus.NewRegistry()` (no global default registerer, `/metrics` served via `promhttp.HandlerFor`), so no cross-registration panic risk. All labels are bounded: `stream`(2), `result`(2), `disp`(3), `op`(4), `reason` (closed const sets), `shard`/`partition` (operator-bounded by table/partition count). No high-cardinality or secret-bearing labels.
- **Concurrency — clean:** prometheus vec/histogram types are goroutine-safe; the producer's per-tick `Sink` factory shares one `*Metrics` safely, and the single-tick CAS plus the renew goroutine join-before-release ordering are preserved. obs `ready` is `atomic.Bool`.

## 2. Code quality findings (all non-blocking)

### P2 — `produced_total{stream=dlq}` and `dlq_total` diverge on a mid-tick abort
`internal/producer/etl.go` (`MarkDLQ` per-chunk at ~190-194 vs `AddProduced` per-tick at ~134-135)

`dlq_total{reason}` is incremented per-chunk immediately after `ProduceDLQ` confirms, but `produced_total` is incremented once at end-of-tick via `AddProduced(totalMain, totalDLQ)`. If the tick aborts mid-way — the per-chunk `ctx.Err()` check fires on lock-loss/SIGTERM, or a later chunk's `ProduceBatch`/cursor-advance errors — `runTick` returns before reaching `AddProduced`. Messages that were genuinely written to Kafka then get counted in `dlq_total` but never in `produced_total{stream=dlq}` (and the chunk's main rows are also dropped from `produced_total{stream=main}`), so the two permanently disagree and `produced_total` under-reports vs its Help text. Data integrity is unaffected (message_id idempotency + CAS). Lowest-risk fix: move the produced accounting per-chunk alongside `MarkDLQ` so both share one increment site, or document that `dlq_total` counts confirmed DLQ writes while `produced_total` counts only fully-completed ticks.

### P2 — `dlq_hard_stop_total` false-positives on graceful shutdown during a DLQ retry
`internal/consumer/consumer.go` `routePoison` (~296-300), with `dlq.go` (~103-110) and `service.go` (~59-63)

`routePoison` treats *any* non-nil error from `dlq.Send` as a hard stop and fires `alert("dlq_hard_stop")` → `MarkDLQHardStop()`. But `Send` returns a bare `ctx.Err()` (not `errDLQHardStop`) when the context is cancelled at the top of the retry loop or during backoff sleep — i.e. on normal SIGTERM while a poison message is mid-retry. That increments the highest-priority page-worthy counter on a graceful shutdown, even when `SpillDir` is configured and no terminal escape occurred. The trigger window is narrow (a poison message actively mid-`Send`-retry exactly when ctx cancels), so spurious counts are bounded. Suggested fix: in `routePoison`, skip the `dlq_hard_stop` alert/count when `errors.Is(err, context.Canceled) || errors.Is(err, context.DeadlineExceeded)`.

### nit — `produce_errors_total` counts marshal failures, not only Kafka write failures
`internal/producer/kafka.go` `ProduceBatch` (~85-92)

`ProduceBatch` calls `MarkProduceError()` on any error from `produce()`, but `produce()` can return a `json.Marshal` error before any `WriteMessages` call, so an encoding bug counts as a "produce error" — contradicting the Help/comment ("Kafka WriteMessages failures") and asymmetric with `ProduceDLQ`, which only counts on the write branch. Effectively unreachable at runtime for plain contract structs; tighten by pushing the counter into `produce()`'s write branch or distinguishing marshal vs write errors.

### nit — `disposition_total{disp="transient"}` is an occurrence counter, Help reads per-item
`internal/consumer/consumer.go` (default branch ~270-272)

`resolvePass` runs once per retry attempt; terminal states (`ok`/`dlq`) are guarded to mark once per item, but the transient branch marks on every failed attempt, so `disposition_total{disp="transient"}` counts retries, not final per-item state. This matches the plan ("每条定 disp 时 Inc()") and is arguably the more useful retry-rate signal, but the Help string "Per-item disposition" overstates it. Either tighten the Help wording or accept it as an occurrence counter.

---

## Coverage notes / what was not verified
- Prometheus scrape config, dashboards, and alert rules live in the deployment repo (plan §4) and are out of scope here; downstream alerting should be aware of the `dlq_hard_stop_total` shutdown false-positive (P2 #2) until fixed.
- Runtime histogram behavior under real Kafka/ES load was not exercised; assertions are on registry/series shape and code paths.
- An additional cross-model advisory pass was attempted but timed out and is therefore absent from this review; the conclusions above rest on the local toolchain run plus the four-dimension adversarial verification.

OctoBoooot (commented):
## Delta re-review (`23af214f` → `a29e074e`) — #39

**Verdict:** Comment

The CI Lint blocker from my prior review is **fixed**. The only remaining red check is `check-sprint`, a maintainer-only governance gate (not code) — so I can't post APPROVE while a required check is red, but on the code merits this is ready. One carried-forward minor remains (author's call).

### ✓ Prior blocker resolved — Lint is green

`internal/consumer/metrics_test.go:139` — the `errcheck check-blank` failure is fixed: `_ = p.processBatch(...)` is now `if err := p.processBatch(...); err != nil { t.Logf(...) }`, which consumes the error via a checked branch. Confirmed live: **Lint = SUCCESS**, plus Build / Test / Vet / CodeQL all green on `a29e074e`. The fix is appropriately scoped — the test asserts metric semantics, so logging the (expected, spill-dependent) error rather than failing on it is the right call.

### Remaining

- **CI: `check-sprint` still FAILURE** — this is the governance gate (linked issue #38 not attached to a sprint on the Octo Board). It needs a maintainer, not a code change. Flagging it as the one thing standing between this and a green merge; it's why this verdict is COMMENT, not APPROVE.
- **Minor (carried forward, unaddressed): `disposition_total{disp="transient"}` counts per retry-attempt** — `consumer.go:272` still `markDisposition(dispTransient)` in the per-pass `default` branch, so a transient item re-counts each retry while `ok`/`dlq` count once per item. Observability-only; legitimately the author's call to defer, but noting it persists. Suggested fix unchanged (mark transient once on entry, or split to `transient_retries_total`).

### Praise

- Tight, surgical fix — a single test file touched (+5/-1) to clear the exact lint diagnostic, with a comment explaining why the error is logged rather than asserted (the no-spill hard-stop path legitimately returns one). No scope creep onto the green-on-merits metrics code.

Once a maintainer attaches #38 to a sprint, this is clear to merge from my side — the code blocker is gone and only the carried-forward observability minor (non-blocking) remains.

Jerry-Xin (approved):
Summary: The PR is in scope for `octo-search-indexer` and cleanly adds Prometheus observability without changing the producer/consumer correctness paths.

💬 Non-blocking

🟡 Warning — `cmd/es-indexer/main.go:74-77` marks `/readyz` ready before `svc.Run` executes startup checks in `internal/consumer/service.go:151-160`. This can briefly report ready even if index/mapping validation is about to fail. Consider moving readiness after those checks or wiring a real ready check.

🟡 Warning — `internal/producer/metrics.go:101-108` and similar `CounterVec` metrics do not pre-initialize known label values. Compared with the old rendered producer metrics, series like `searchetl_producer_produced_total{stream="main"}` may be absent until nonzero. That can surprise dashboards/alerts expecting zero-valued series. Consider `WithLabelValues(...).Add(0)` for known labels.

🔵 Suggestion — `internal/consumer/metrics.go:20-27` documents seven consumer series, but `indexer_dlq_write_exhausted_total` is also registered at `internal/consumer/metrics.go:61-65`. Either include it in the public metric list/plan or drop it if it is not intended as part of the contract.

🔵 Suggestion — `internal/producer/obs.go:22-24` still says the obs server uses only `net/http` and adds no dependency, but the implementation now uses `promhttp`. Worth updating the comment.

✅ Highlights

- Private Prometheus registries per binary are a good fit and avoid global registration collisions.
- Producer instrumentation is placed after confirmed DLQ writes and does not alter cursor advancement semantics.
- Consumer DLQ hard-stop vs write-exhausted metrics are separated clearly.
- Verified locally: `go test ./internal/producer/... ./internal/consumer/...`, `go test ./...`, and `go vet ./...` all pass.

mochashanyao (approved):
[Octo-Q · automated review]

---

# octo-search-indexer PR#39 Review Report
**PR:** [feat(obs): add Prometheus metrics for producer and consumer](https://github.com/Mininglamp-OSS/octo-search-indexer/pull/39)
**Head SHA:** `a29e074ed6fd37e87af0c0a1e8fa7aa1ba79fadc`
**Scope:** 21 files, +1085/-122 lines — Prometheus SDK migration (producer) + greenfield metrics/obs (consumer)

---

## 1. Verification Summary

| Check | Result | Evidence |
|---|---|---|
| `go build ./...` | ✅ | exit 0, clean |
| `go vet ./...` | ✅ | exit 0, clean |
| `go test ./internal/producer/...` | ✅ | 0.176s, all pass |
| `go test ./internal/consumer/...` | ✅ | 0.499s, all pass |
| Private registry isolation | ✅ | Both legs use `prometheus.NewRegistry()`, no global default — cross-registration impossible |
| Nil-guard coverage | ✅ | Every metrics call-site guarded; production path always non-nil via `NewMetrics()` |
| Data-flow integrity | ✅ | All consumed data traced to upstream constructors; no stale/empty/short-circuited paths |
| Thread safety | ✅ | Prometheus primitives internally atomic; `*Metrics` pointer set once at construction, never mutated |
| Correctness paths untouched | ✅ | Transaction boundaries, cursor CAS, DLQ decisions, offset commit logic — zero diff |

---

## 2. Findings

### F1 — Consumer `disposition_total` double-counts transient items across retries (P2 / note)

**Severity:** P2 (informational)
**Diff-scope:** New behavior introduced by this PR (consumer metrics are greenfield).

In `processBatch` (`consumer.go:130-146`), the retry loop calls `resolvePass` repeatedly. Each call to `resolvePass` invokes `markDisposition(dispTransient)` for every item still transient — meaning a single message that takes 3 retry rounds increments `disposition_total{disp="transient"}` **3 times**.

```
// consumer.go processBatch loop:
for attempt := 0; ; attempt++ {
    changed, err := p.resolvePass(...)  // marks transient for each unresolved item
    ...
}
```

**Impact:** The counter reflects "per-resolve-pass outcomes" not "unique messages". `rate(disposition_total{disp="transient"})` will overcount relative to `rate(disposition_total{disp="ok"})` under transient-heavy loads. This is not a bug per se (it's useful for measuring retry pressure), but it should be documented to prevent dashboard misinterpretation.

**Suggestion:** Add a one-line comment on the `disposition_total` help text or plan doc clarifying the counter reflects per-pass outcomes, not unique message counts.

### F2 — Consumer obs `readyz` set to true before ES/Kafka validation (P2)

**Severity:** P2 (should-fix before merge)
**Diff-scope:** New behavior introduced by this PR.

In `cmd/es-indexer/main.go:78-79`:
```go
obs.SetReady(true)    // ← before svc.Run(ctx)
...
return svc.Run(ctx)   // ← EnsureIndex + AssertLiveMappingCompatible here
```

The consumer claims readiness before `svc.Run` performs `EnsureIndex` and `AssertLiveMappingCompatible`. If ES is unreachable or mapping-incompatible, the pod reports ready to the orchestrator but immediately returns a fatal error, causing a crashloop that the orchestrator won't distinguish from a genuine liveness failure.

**Suggestion:** Set `obs.SetReady(true)` inside `svc.Run` after the ES assertions pass (or after the first successful `fetchBatch`), mirroring the producer side which has a real `ReadyCheck` function.

### F3 — Both binaries default obs to `:9090` (P2 / nit)

**Severity:** P2 (nit)
**Diff-scope:** New — consumer obs server is introduced by this PR.

Both `INDEXER_OBS_ADDR` and `PRODUCER_OBS_ADDR` default to `:9090`. While these typically run in separate pods/containers, a same-host deployment (e.g. docker-compose, dev environment) would see a port collision on the second binary to start.

**Suggestion:** Default consumer to `:9091` to avoid same-host collision. Low urgency — K8s deployments are unaffected.

---

## 3. Suggestions

1. **F1 doc fix:** Add a comment in `metrics.go` or the plan doc: `// disposition_total counts per-resolve-pass outcomes, not unique messages; transient items are counted once per retry round.`

2. **F2 readyz fix:** Move `obs.SetReady(true)` to after `svc.Run`'s ES validation, or wire a `ReadyCheck` function that pings ES reachability.

3. **F3 port default:** Consider `:9091` for consumer default. Optional.

4. **Missing alerter event mapping (nit):** `metricsAlerter.Alert` in `service.go` does not map `"dlq_marshal_failed"` or `"dlq_spilled_to_disk"` events to any counter. These events are log-only. If operational visibility into these is desired, consider adding counters. Low priority — these are edge-of-edge events.

---

## 4. Additional Findings

- **Producer `AddProduced` zero-skip:** When a tick produces zero messages (both main=0 and dlq=0), no `produced_total` sample is emitted. Dashboards using `rate(produced_total[5m])` will see gaps during quiet periods rather than a flat zero line. This is standard Prometheus counter behavior but worth noting for SLO dashboard design.

- **`NewKafkaProducer` is dead code in production path:** `cmd/searchetl-producer/main.go` always calls `NewKafkaProducerWithMetrics`. The no-metrics `NewKafkaProducer` wrapper is never used by the production binary. Retained for test/convenience use — acceptable.

- **Producer `produce_errors_total` counts batch failures, not message failures:** A single `WriteMessages` call with 100 messages counts as 1 error. Consistent with the DLQ produce error counting pattern. Document accordingly.

---

## 5. Data-Flow Tracing

### Producer metrics flow (all verified)

```
main.go: metrics := producer.NewMetrics()
  → ETL.metrics → runLocked(..., metrics) → renewUntilDone(..., metrics) → MarkLockRenewFailure [nil-guarded]
  → ETL.metrics → runTick → runChunk(..., metrics) → ObserveReadBatch [nil-guarded, unconditional after ReadStableBatchTx]
  → ETL.metrics → runTick → runChunk → MarkDLQ(plan.dlq[i].Reason) [nil-guarded, only after successful ProduceDLQ]
  → ETL.metrics → runTick → SetCursor [nil-guarded, gated on plan.advanced]
  → ETL.metrics → runTick → AddProduced [nil-guarded, after full table loop]
  → Scheduler.metrics → tick → ObserveTickDuration + MarkTick("ok"/"error") [nil-guarded, unconditional]
  → KafkaProducer.metrics → ProduceBatch/ProduceDLQ → MarkProduceError [nil-guarded, on WriteMessages err]
  → ObsServer → promhttp.HandlerFor(metrics.Registry()) [nil-guarded at construction]
```

### Consumer metrics flow (all verified)

```
service.go: metrics := NewMetrics()
  → Processor.metrics → resolvePass → markDisposition [nil-guarded, per-item per-pass]
  → Processor.metrics → resolvePass → ObserveIO("es_bulk") [nil-guarded, unconditional]
  → Processor.metrics → resolvePass → MarkIOError("es_bulk") [nil-guarded, on bulkErr]
  → Processor.metrics → routePoison → MarkDLQ(reason) [nil-guarded, after successful DLQ send]
  → Processor.metrics → commitPrefixes → ObserveIO("kafka_commit") + SetCommittedOffset [nil-guarded, after success]
  → kafkaSource.metrics → Fetch → ObserveIO("kafka_fetch") [unconditional] + MarkIOError [gated: ctx.Err()==nil]
  → kafkaDLQSink.metrics → WriteDLQ → ObserveIO("dlq_send") + MarkIOError [nil-guarded]
  → metricsAlerter.metrics → Alert → MarkBulkError/MarkDLQHardStop/MarkDLQWriteExhausted [nil-guarded]
  → ObsServer → promhttp.HandlerFor(metrics.Registry()) [nil-guarded at construction]
```

**No data-flow gaps found.** Every consumed metric pointer is traced to a single `NewMetrics()` constructor call. No stale, empty, or short-circuited paths detected.

---

## 6. Blind-Point Checklist (R5)

- **C1 — Dual-path parity:** N/A. This PR adds observability instrumentation; no symmetric add/remove/subscribe/unsubscribe paths are touched.

- **C2 — Control-flow ordering / nested reuse:** Clear. Metrics methods are simple delegates to prometheus primitives (internally thread-safe). No nested/compound calls that could double-act. Nil-guard pattern is uniform across all call-sites.

- **C3 — Authorization boundary ≠ capability boundary:** N/A. No authorization changes. The obs server endpoints are internal observability — no user-facing auth boundary.

- **C4 — Authorization lifecycle / container-member state cascade:** N/A. No authentication or authorization changes in this PR.

- **C5 — Build ≠ runtime path correctness:** Clear. `go build`, `go vet`, and `go test` all pass. The prometheus SDK is used via standard `promhttp.HandlerFor` — no relative paths, no extension-specific concerns, no heredoc issues. The two binaries run in separate containers in production; default port collision (F3) is a deployment convenience concern, not a runtime correctness issue.

- **C6 — Governance/policy document self-consistency:** N/A. No governance, security, or policy documents are modified.

---

## 7. Cross-Round Blocker Re-check (R6)

---

## Verdict

**All findings are P2/nit.** No P0 or P1 issues found. The PR is a clean observability instrumentation that correctly avoids touching any correctness paths (transaction boundaries, cursor CAS, DLQ decisions, offset commit logic are all untouched as claimed). Data-flow tracing confirms all consumed metrics pointers are properly sourced and nil-guarded. Build, vet, and tests all pass.

F2 (premature readyz) is the most actionable item and should be a straightforward fix before merge.

[Octo-Q] verdict: APPROVE

yujiawei (approved):
# Code Review — PR #39 (octo-search-indexer)

**Scope:** observability-only — Prometheus metrics for the `searchetl-producer` (SDK migration + 2 histograms + 3 counters) and the `es-indexer` consumer (metrics + obs server built from scratch, `logAlerter` → `metricsAlerter`). Reviewed against base `origin/main` (merge-base `7fedf7d`).

## 1. Spec compliance

**Spec: ✅**

The stated constraint is *instrumentation + obs infra only — do not touch correctness paths (transaction boundaries, cursor CAS, DLQ decisions, offset commit, transient-retry convergence)*. The diff honors it:

- **No correctness-path change.** Every signature change (`runChunk +metrics`, `runLocked`/`renewUntilDone +metrics`, `NewProcessor +metrics`, `newKafkaSource`/`newKafkaDLQSink +metrics`, `NewKafkaProducerWithMetrics`) only threads a `*Metrics` and adds `Inc`/`Observe`/`Set` calls beside existing decision points. SQL, `FOR UPDATE`/commit boundaries, cursor CAS (`AdvanceCursor`), offset-commit prefix logic, and the transient-retry loop are byte-identical.
- The two reorderings are behavior-preserving: `scheduler.tick()` now calls `tickFn` once and logs iff `err != nil` (same message, same cadence); `ProduceBatch` one-liner → `if/return` returns the same error, only counting on the error branch.
- `planChunk` stays pure — `dlq_total{reason}` is counted in `runChunk` *after* `ProduceDLQ` confirms, not inside the planner.
- **Missing:** none. **Extra-built:** one extra counter beyond the linked issue/plan's stated "7 consumer series" — `dlq_write_exhausted_total` (8 total). This is a justified review-driven addition (it disambiguates write-exhaustion from hard-stop to prevent double-counting) and is pure observability, so it does not violate the instrumentation-only scope — but the plan/PR text says 7 while the code registers 8; reconcile the count. **Deviation:** none.

Build/vet/test verified locally: `go build ./...`, `go vet ./...`, `go test ./...` all green; `go test -race ./internal/consumer/... ./internal/producer/...` green; `gofmt` clean on all touched files; `go mod verify` passes and `go mod tidy` is a no-op (all dep changes are forward upgrades pulled by client_golang v1.23.2, no downgrades).

## 2. Code quality

**Quality: Approved**

No P0/P1 issues. The metric core is correct: the `dlq_hard_stop_total` vs `dlq_write_exhausted_total` split is right (hard-stop counted once in the no-spill escape; exhausted counted on both spill and hard-stop paths), registries are private per-binary with a single `MustRegister`, all labels are bounded enums/config values (no cardinality-explosion vector), and obs servers are shut down on both the normal and idle paths.

The following are **P2 / advisory** — none blocks merge:

- **P2 — `dlq_hard_stop` event fires for *any* `dlq.Send` error, including `context.Canceled`.** `internal/consumer/consumer.go:296-299` maps every `dlq.Send` error to the `dlq_hard_stop` alert, and `dlq.Send` returns `ctx.Err()` on shutdown/backoff cancellation (`internal/consumer/dlq.go:103-105,108-110`). Since `metricsAlerter` maps that event to `MarkDLQHardStop()` (`service.go:59-63`), a normal SIGTERM mid-DLQ-write can emit a spurious `dlq_hard_stop_total` increment — a false "terminal escape" signal with no actual retry exhaustion. Consider skipping the counter (and/or the alert) when `errors.Is(err, context.Canceled)`.
- **P2 — `disposition_total{disp="transient"}` is counted per-retry-round, not per-item.** `internal/consumer/consumer.go:270-272`: an item that stays transient is re-marked every resolve round, while `ok`/`dlq` are marked once per item. So `transient` is an attempt-count and `ok`/`dlq` are item-counts — `transient/(ok+dlq+transient)` over-reports under sustained 429/5xx backpressure. The struct/Help text says "per-item", which conflicts. Either mark transient only on first transition, or reword the Help to "per resolve-pass attempt" and add a separate terminal-disposition counter for dashboards.
- **P2 — `dlq_total{reason}` is not incremented on the hard-stop escape path.** `routePoison` (`consumer.go:301-303`) only counts `dlq_total` when `dlq.Send` succeeds; on hard-stop the message was dead-lettering but is not counted under `dlq_total`. This is defensible (it never reached the DLQ topic, and `dlq_hard_stop_total` covers the gap) but worth a one-line doc note so `dlq_total` isn't read as "all attempted dead-letters".
- **P2 — consumer `/readyz` flips ready before fail-closed startup gates run.** `cmd/es-indexer/main.go:74-87` calls `SetReady(true)` before `svc.Run` executes `EnsureIndex` / `AssertLiveMappingCompatible` (`service.go:144-161`), and passes `ready=nil` (no dependency probe), diverging from the producer leg which uses a real DB+Redis `ReadyCheck` and sets ready only after start-up. The plan explicitly defers the deep ping to a later phase, so `ready=nil` is sanctioned; the ordering (ready-before-gates) is the only nit — a pod destined to crash on a bad index briefly advertises 200. Move `SetReady(true)` after the gates to match the producer.
- **P2 — both binaries default obs to `:9090` and `ObsServer.Start` swallows bind errors.** `cmd/es-indexer/main.go:51` / `internal/producer/config.go` both default `:9090`; `obs.go` `Start` only logs a non-`ErrServerClosed` `ListenAndServe` error. In the normal one-deployment-per-binary topology there's no collision, but if co-located (same pod/sidecar/local dev) the second leg silently loses `/metrics`+`/healthz`+`/readyz` while appearing healthy. Operationally mitigable via env override; optionally surface a fatal bind error on the non-idle path.
- **P2 — doc/count reconciliation.** As noted above, the PR body, plan §3.2, and the `metrics.go` struct comment all say 7 consumer series while 8 are registered (`dlq_write_exhausted_total` is live and tested). Update the text to 8.

Test quality is good: `TestProcessBatch_DispositionMetrics` and `TestDLQHardStopVsExhausted_MetricSemantics` drive the real `processBatch` path (not just the metric helpers) and lock in the hard-stop/exhausted semantics.

## 3. Overall verdict

**APPROVE** — spec ✅ (instrumentation-only honored, no correctness path touched) and quality Approved (no P0/P1). The six P2 items above are non-blocking polish; the `context.Canceled` → false `dlq_hard_stop_total` one is the most worth a quick follow-up since it can generate spurious alerts on every graceful shutdown.

## 4. Suggestions

- Guard the `dlq_hard_stop` alert/counter with `if !errors.Is(err, context.Canceled)`.
- Clarify `disposition_total` semantics (per-attempt vs per-item) in the Help text or emission site.
- Reconcile the "7 vs 8 consumer series" count across PR body, plan, and the `metrics.go` struct doc.
- Move consumer `SetReady(true)` to after the startup gates to mirror the producer leg.

PR #47, opened 2026-07-06
feat: file content indexing (payload.file.content + file-extractor + backfill)

aunknown:
## What

End-to-end file content search for octo message payload.type=8 (files).

Extract text from PDF/Office/etc files via Apache Tika, index into OpenSearch
under payload.file.content (text, ik_max_word analyzer), and expose through
backfill (one-shot Job for existing files) and the message-consumer pipeline
(new file-extractor service for incoming messages).

## Why

Product requirement (PRD `Octo搜索需求文档 v1.0`): users need to search
for keywords within file contents (not just filenames) and locate the
containing message. Current search only indexes message text; file bodies
are opaque.

## Design

See `docs/file-content-indexing-feasibility.md` and
`docs/file-content-indexing-implementation.md` for full technical design
(architecture, tool selection, DLQ topology, safety controls).

## Testing

- Unit + integration tests for all new packages (fileextract, filebackfill,
  esindex partial-update)
- Race-tested (`go test -race`)
- Mapping-compat startup fail-closed gate: file-extractor refuses to start
  if live OS mapping lacks required fields or contains forbidden
  `_source.excludes` over preserved paths
- Post-merge acceptance: manual real-OS validation in test env before prod
  rollout (validates scripted_upsert preserve, tombstone marker + backfill
  exclude, deployment-order gate)

## Deployment

**Requires one-time admin action before rolling pods**:

```
PUT /<current-backing-index>/_mapping
{
  "_source": { "excludes": [] },
  "properties": {
    "payload": { "properties": { "file": { "properties": {
      "contentMeta": { "properties": {
        "status": { "type": "keyword", "ignore_above": 32 },
        "reason": { "type": "keyword", "ignore_above": 64 }
      }}
    }}}}
  }
}
```

Then rollout es-indexer + file-extractor Deployments. If mapping PUT is
skipped, pods will CrashLoop on startup (mapping-compat gate) with a
message pointing to this runbook.

Full runbook + tombstone semantics in `internal/esindex/mapping/README.md`.

## Supersedes

Consolidation of #46 (25+ reviews / 5 review rounds). Prior discussion
remains as reference. All previously-flagged blocking findings are addressed
in this consolidated commit:

- SSRF protection on file downloader
- `_source.excludes` compatibility with `scripted_upsert` preserve
- 409 version-conflict classified transient (matches sibling writer)
- Backfill tombstone marker prevents unbounded DLQ growth on rerun
- Tombstone whitelist distinguishes permanent-fail vs transient-nature reasons
- mapping-compat startup gate prevents deployment-order errors

Deferred to follow-up PRs (tracked in workspace):
- OSPermanent alerter (observability, low severity)
- CDN 404 split from download_failed (backfill efficiency, bounded impact)
- sourceExcludesContains wildcard defense-in-depth

/cc @yujiawei @mochashanyao @Jerry-Xin @OctoBoooot

### DIFF
--- a/cmd/file-content-backfill/main.go
+++ b/cmd/file-content-backfill/main.go
@@ -1,0 +1,176 @@
+// Command file-content-backfill 一次性 K8s Job：把 IDX-4 file-extractor 上线**之前**的历史
+// type=8 文件消息回填 payload.file.content（v1.12 file content indexing）。
+//
+// 数据链路：
+//
+//	OpenSearch → scroll query [payload.type=8 且 must_not exists payload.file.content]
+//	           → 逐条限速抽取（复用 internal/fileextract.Extractor）
+//	           → OS partial `_update` 只写 payload.file.content + contentMeta
+//
+// 幂等：query 的 `must_not exists content` 天然跳过已抽取 doc，中断重跑无副作用。
+//
+// 用法（K8s Job）：
+//
+//	file-content-backfill \
+//	  -es https://10.10.148.6:9200,https://10.10.148.15:9200,https://10.10.148.12:9200 \
+//	  -es-user octosearch -es-pass "$OS_PASSWORD" \
+//	  -es-index octo-message \
+//	  -tika-url http://tika-service:9998 \
+//	  -rate 50 -timeout 2h
+//
+// 退出码：Stats.OSTransient > 0 或 DLQ 占比 > 10% → exit 1 让 K8s 报错；正常 exit 0。
+//
+// 🔴 隔离纪律：真实环境执行须 Yu / 运维 sign-off，主人决策 backfill 走 K8s Job 一次性跑
+// （不常驻），跑完手动 delete Job 释放资源。
+package main
+
+import (
+	"context"
+	"flag"
+	"fmt"
+	"log"
+	"os"
+	"os/signal"
+	"strings"
+	"syscall"
+	"time"
+
+	fbf "github.com/Mininglamp-OSS/octo-search-indexer/internal/filebackfill"
+)
+
+func main() {
+	log.SetFlags(log.LstdFlags | log.LUTC)
+	if err := run(); err != nil {
+		log.Printf("file-content-backfill error: %v", err)
+		os.Exit(1)
+	}
+}
+
+func run() error {
+	var (
+		esAddrs     = flag.String("es", envOr("BACKFILL_ES", ""), "comma-separated OpenSearch addresses (required)")
+		esIndex     = flag.String("es-index", envOr("BACKFILL_ES_INDEX", "octo-message"), "OpenSearch index / alias")
+		esUser      = flag.String("es-user", os.Getenv("BACKFILL_ES_USER"), "OS username (optional)")
+		esPass      = flag.String("es-pass", os.Getenv("BACKFILL_ES_PASS"), "OS password (optional)")
+		tikaURL     = flag.String("tika-url", envOr("BACKFILL_TIKA_URL", "http://tika-service:9998"), "Tika Server URL")
+		rate        = flag.Float64("rate", envFloat("BACKFILL_RATE", 50), "extraction rate limit (docs/s; <=0 = unlimited)")
+		scrollSize  = flag.Int("scroll-size", envInt("BACKFILL_SCROLL_SIZE", 500), "OS scroll batch size")
+		scrollTTL   = flag.Duration("scroll-ttl", envDuration("BACKFILL_SCROLL_TTL", 5*time.Minute), "scroll keep-alive TTL")
+		timeout     = flag.Duration("timeout", envDuration("BACKFILL_TIMEOUT", 2*time.Hour), "overall Job timeout")
+		downloadMS  = flag.Int("download-timeout-ms", envInt("BACKFILL_DOWNLOAD_TIMEOUT_MS", 30000), "CDN download timeout (ms)")
+		extractMS   = flag.Int("extract-timeout-ms", envInt("BACKFILL_EXTRACT_TIMEOUT_MS", 30000), "Tika extract timeout (ms)")
+		maxSize     = flag.Int64("max-file-size", int64(envInt("BACKFILL_MAX_FILE_SIZE_BYTES", 20*1024*1024)), "max file size (bytes)")
+		maxContent  = flag.Int("max-content-bytes", envInt("BACKFILL_MAX_CONTENT_BYTES", 256*1024), "max extracted content bytes")
+		httpRetries = flag.Int("http-retries", envInt("BACKFILL_HTTP_RETRIES", 3), "CDN GET retry count")
+		dlqRatioMax = flag.Float64("max-dlq-ratio", 0.10, "max allowed DLQ/scanned ratio; over threshold → exit 1")
+	)
+	flag.Parse()
+
+	if *esAddrs == "" {
+		return fmt.Errorf("es addresses required (-es or BACKFILL_ES)")
+	}
+
+	cfg := fbf.Config{
+		ESAddresses:     splitCSV(*esAddrs),
+		ESIndex:         *esIndex,
+		ESUsername:      *esUser,
+		ESPassword:      *esPass,
+		TikaURL:         *tikaURL,
+		DownloadTimeout: time.Duration(*downloadMS) * time.Millisecond,
+		ExtractTimeout:  time.Duration(*extractMS) * time.Millisecond,
+		MaxFileSize:     *maxSize,
+		MaxContentBytes: *maxContent,
+		HTTPRetries:     *httpRetries,
+		Rate:            *rate,
+		ScrollSize:      *scrollSize,
+		ScrollTTL:       *scrollTTL,
+		Timeout:         *timeout,
+	}
+
+	runner, err := fbf.NewRunner(cfg)
+	if err != nil {
+		return fmt.Errorf("build runner: %w", err)
+	}
+
+	ctx, cancel := context.WithTimeout(context.Background(), *timeout)
+	defer cancel()
+	ctx, stop := signal.NotifyContext(ctx, syscall.SIGINT, syscall.SIGTERM)
+	defer stop()
+
+	log.Printf("file-content-backfill starting: es_index=%s tika=%s rate=%.1f scroll_size=%d timeout=%v",
+		*esIndex, *tikaURL, *rate, *scrollSize, *timeout)
+
+	stats, err := runner.Run(ctx)
+	if err != nil {
+		return fmt.Errorf("runner.Run: %w (stats: %+v)", err, stats)
+	}
+
+	// 退出码判定：OSTransient>0 或 DLQ 占比过高 → 让 K8s Job 报错
+	if stats.OSTransient > 0 {
+		return fmt.Errorf("os transient errors during backfill: %d (stats: %+v)", stats.OSTransient, stats)
+	}
+	if stats.Scanned > 0 {
+		ratio := float64(stats.DLQ) / float64(stats.Scanned)
+		if ratio > *dlqRatioMax {
+			return fmt.Errorf("dlq ratio %.3f > threshold %.3f (dlq=%d scanned=%d)", ratio, *dlqRatioMax, stats.DLQ, stats.Scanned)
+		}
+	}
+	return nil
+}
+
+func envOr(key, def string) string {
+	if v := os.Getenv(key); v != "" {
+		return v
+	}
+	return def
+}
+
+func envInt(key string, def int) int {
+	v := os.Getenv(key)
+	if v == "" {
+		return def
+	}
+	var n int
+	if _, err := fmt.Sscanf(v, "%d", &n); err != nil {
+		log.Printf("file-content-backfill: invalid %s=%q, using default %d", key, v, def)
+		return def
+	}
+	return n
+}
+
+func envFloat(key string, def float64) float64 {
+	v := os.Getenv(key)
+	if v == "" {
+		return def
+	}
+	var f float64
+	if _, err := fmt.Sscanf(v, "%f", &f); err != nil {
+		log.Printf("file-content-backfill: invalid %s=%q, using default %f", key, v, def)
+		return def
+	}
+	return f
+}
+
+func envDuration(key string, def time.Duration) time.Duration {
+	v := os.Getenv(key)
+	if v == "" {
+		return def
+	}
+	d, err := time.ParseDuration(v)
+	if err != nil {
+		log.Printf("file-content-backfill: invalid %s=%q, using default %v", key, v, def)
+		return def
+	}
+	return d
+}
+
+func splitCSV(s string) []string {
+	parts := strings.Split(s, ",")
+	out := make([]string, 0, len(parts))
+	for _, p := range parts {
+		if t := strings.TrimSpace(p); t != "" {
+			out = append(out, t)
+		}
+	}
+	return out
+}
--- a/cmd/file-content-backfill/main_test.go
+++ b/cmd/file-content-backfill/main_test.go
@@ -1,0 +1,74 @@
+package main
+
+import (
+	"os"
+	"strings"
+	"testing"
+)
+
+// resetBackfillEnv 清空 BACKFILL_ 前缀所有 env，避免 test 污染。
+func resetBackfillEnv(t *testing.T) {
+	t.Helper()
+	for _, kv := range os.Environ() {
+		if idx := strings.IndexByte(kv, '='); idx > 0 {
+			key := kv[:idx]
+			if strings.HasPrefix(key, "BACKFILL_") {
+				t.Setenv(key, "")
+			}
+		}
+	}
+}
+
+// TestSplitCSV 覆盖 CSV 解析各种输入形态。
+func TestSplitCSV(t *testing.T) {
+	cases := []struct {
+		in   string
+		want []string
+	}{
+		{"", []string{}},
+		{"a", []string{"a"}},
+		{"a,b,c", []string{"a", "b", "c"}},
+		{"a, b , c", []string{"a", "b", "c"}}, // trim
+		{",a,,b,", []string{"a", "b"}},        // 忽略空
+	}
+	for _, c := range cases {
+		got := splitCSV(c.in)
+		if len(got) != len(c.want) {
+			t.Errorf("splitCSV(%q): len got=%d want=%d (%v)", c.in, len(got), len(c.want), got)
+			continue
+		}
+		for i := range got {
+			if got[i] != c.want[i] {
+				t.Errorf("splitCSV(%q)[%d]: got %q want %q", c.in, i, got[i], c.want[i])
+			}
+		}
+	}
+}
+
+// TestEnvOr 默认值 / env 覆盖。
+func TestEnvOr(t *testing.T) {
+	resetBackfillEnv(t)
+	if got := envOr("BACKFILL_NOT_SET_XYZ", "default"); got != "default" {
+		t.Errorf("envOr default: got %q", got)
+	}
+	t.Setenv("BACKFILL_TEST_KEY", "override")
+	if got := envOr("BACKFILL_TEST_KEY", "default"); got != "override" {
+		t.Errorf("envOr override: got %q", got)
+	}
+}
+
+// TestEnvInt 有效整数 / 无效字符串走 default / env 缺省。
+func TestEnvInt(t *testing.T) {
+	resetBackfillEnv(t)
+	if got := envInt("BACKFILL_NOT_SET", 42); got != 42 {
+		t.Errorf("envInt default: got %d", got)
+	}
+	t.Setenv("BACKFILL_NUM", "100")
+	if got := envInt("BACKFILL_NUM", 42); got != 100 {
+		t.Errorf("envInt override: got %d", got)
+	}
+	t.Setenv("BACKFILL_BAD", "not-a-number")
+	if got := envInt("BACKFILL_BAD", 42); got != 42 {
+		t.Errorf("envInt invalid falls back to default: got %d", got)
+	}
+}
--- a/cmd/file-extractor/main.go
+++ b/cmd/file-extractor/main.go
@@ -1,0 +1,183 @@
+// Command file-extractor 是 octo-im 消息检索管线的文件正文抽取服务（v1.12）。
+// 消费与 es-indexer 同一个 Kafka topic `octo.message.v1{,.prod}`，独立 consumer group
+// `file-extractor`（不抢 es-indexer 位点）。命中 payload.type=8 (File) 时下载 CDN 文件 → 调
+// Tika HTTP 抽取正文 → OS partial `_update` 只写 payload.file.content + payload.file.contentMeta。
+// 命中非 file 类型 → commit 位点跳过。
+//
+// 与 es-indexer 的分工：
+//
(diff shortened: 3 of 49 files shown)


OctoBoooot (approved):
## Review: file content indexing — payload.file.content + file-extractor + backfill (#47)

**Verdict:** Approve. This is the consolidated resurrection of the closed PR #46 (same author/title), and all five findings resolved across #46's review history are byte-verified intact at this head with no regressions. Code CI is green; only `check-sprint` (governance) is red, so this is code-ready but not auto-mergeable — `mergeStateStatus` stays BLOCKED until the sprint gate clears + human sign-off.
Risk tier: high (consumer/kafka/ratelimit paths). Large change (49 files) → `needs-human-review` applied. Note: the +10594 headline is mostly tests + go.sum; the real non-test code surface is ~14 files / ~1800 lines, all the same modules reviewed across #46.

### Verified intact (each checked against source at this head)
- ✓ **SSRF guard** (`ssrf.go` + `download.go`) — scheme allowlist (https), host allowlist (cdn.deepminer.com.cn), dial-time rejection of loopback/link-local/private/CGNAT/ULA/metadata IPs, and `ssrfCheckRedirect` re-validates on redirect. Wired via `DialContext` + `CheckRedirect`, with a pre-check before the retry loop.
- ✓ **Consumer offset/DLQ** (`decision.go` + `consumer.go`) — `partitionCommitPoints` groups by (topic,partition), sorts by offset, breaks at first transient; bounded in-place retry with `MaxRetriesPerMessage` (default 10) → forced DLQ on exhaustion; permanents routed to DLQ.
- ✓ **Overwrite / Blocker #3** — `octo-message.json` has no `_source.excludes`; `writer.go` uses scripted_upsert + Painless preserve reading `ctx._source.payload.file`; `mapping_compat.go` `forbiddenSourceExcludes` fail-closed startup assertion present.
- ✓ **409 version-conflict** — `writer.go isPermanentStatus` returns false for `http.StatusConflict` (transient), aligned with `oswriter.go classifyOSErr`.
- ✓ **Backfill tombstone** — `extractor.go` writes tombstone on permanent DLQ only; `oswriter.go WriteTombstone` writes `contentMeta.status="unextractable"`; `source.go` backfill query `must_not term contentMeta.status=unextractable`; and critically `octo-message.json` maps `contentMeta.status` as **`keyword`** (`ignore_above:32`) so the term filter matches rather than silently no-op.

### New in #47 (beyond #46 final state) — low severity, no concern
- Tombstone-reason whitelist (`extractor.go tombstoneReasons`): restricts tombstoning to genuinely permanent reasons (blacklist_ext/oversize/encrypted/empty_extract/extract_error) and explicitly excludes transient-natured `download_failed`/`extract_timeout` to preserve backfill retry eligibility. Strictly safer than blanket tombstoning — correct. (Addresses the Round-4 lml2468 note per the PR body.)
- DLQ handler spill-to-disk + hard-stop, backfill bridge/ratelimit plumbing — feature infrastructure, ctx-aware; ratelimiter has a correct `burst<1` guard.

### CI
- ✓ Build / Test / Vet / Lint / CodeQL / secret-scan / osv-scan all green. Only `check-sprint` (governance, maintainer-cleared) red → merge stays BLOCKED on that + the human-review label, not on code.

### Praise
- This closes out a genuinely hard, five-round arc (SSRF → consumer/DLQ → overwrite/#3 → 409 regression → backfill tombstone), and every fix survived into the consolidated PR with its regression test and defense-in-depth guard (the `forbiddenSourceExcludes` boot assertion and the keyword-mapped tombstone filter are the standouts). The one refinement layered on top (transient reasons excluded from tombstoning) shows the author is still reasoning about failure-mode correctness, not just re-landing the diff. Strong close.

Jerry-Xin (approved):
The PR is relevant to `octo-search-indexer` and the main correctness paths look sound: file content extraction is scoped to type 8 messages, OpenSearch writes preserve extractor-owned fields, SSRF controls are present, and `go test ./...` passes.

💬 Non-blocking

- 🟡 Warning: SIGTERM during a backfill extraction can still become a failed Job. `Runner.processOne` counts any extractor error as `OSTransient`, including `context.Canceled`, and `cmd/file-content-backfill` later exits non-zero when `stats.OSTransient > 0`. See `internal/filebackfill/runner.go:180` and `cmd/file-content-backfill/main.go:103`. Consider classifying context cancellation before incrementing `OSTransient`.

- 🟡 Warning: `retry_exhausted` DLQ records omit `messageId` and file context, even though the original message is still available. This makes replay/ops harder than other DLQ reasons. See `internal/fileextract/consumer.go:202` and `internal/fileextract/consumer.go:328`. A fallback to `string(m.Key)` for `MessageID` would improve consistency.

- 🔵 Suggestion: backfill does not expose or forward the file-extractor SSRF allowlist settings. `Config.ToExtractorConfig` only forwards timeout/size/retry settings, so future non-default storage hosts may work in `file-extractor` but fail in `file-content-backfill`. See `internal/filebackfill/config.go:50` and `cmd/file-content-backfill/main.go:50`.

- 🔵 Suggestion: `_source.excludes` compatibility currently checks exact paths only. That catches the known bad config, but wildcard excludes such as `payload.file.*` would still break scripted preserve. See `internal/esindex/mapping_compat.go:155`.

✅ Highlights

- The in-place retry state machine avoids the Kafka offset skip failure mode and handles per-partition commit points correctly.
- The scripted upsert path explicitly preserves `payload.file.content` and `contentMeta`, addressing the dual-writer overwrite risk.
- SSRF protection is layered with URL allowlisting, redirect checks, and dial-time private IP blocking.
- Snowflake `messageId` parsing in backfill uses `json.Decoder.UseNumber`, preserving ID precision.

OctoBoooot (commented):
## Addendum to my review — #47 (reconciling @Jerry-Xin's four 🟡)

My APPROVE stands (all findings below are non-blocking), but @Jerry-Xin raised four 🟡 I verified against source at this head. One is a real gap I missed; the others I'm calibrating with evidence.

### Real, and I missed it — backfill doesn't forward the SSRF allowlist config (minor)
- Confirmed: `cmd/file-extractor/main.go:129-134` reads `ALLOWED_DOWNLOAD_HOSTS`/`ALLOWED_DOWNLOAD_SCHEMES` from env, but `cmd/file-content-backfill/main.go` builds `fbf.Config{}` with no such fields (and `internal/filebackfill/config.go` has no allowlist fields at all).
- **Not a security hole** — it fails *closed*: `ssrf.go:43-44` substitutes `defaultAllowedDownloadHosts = ["cdn.deepminer.com.cn"]` when the list is empty, so the backfill downloader still enforces SSRF. The consequence is operational: a deployment that overrides the allowlist (e.g. an internal COS host) would have file-extractor honor it but the backfill Job silently NOT — backfill downloads from the custom host would be *blocked*, not leaked. So backfill can't be pointed at a non-default CDN without a code change. Correctly 🟡. Fix: thread the same two env vars into `fbf.Config` + the backfill downloader. Credit to Jerry-Xin — I checked the guard existed and was wired in file-extractor but didn't verify the backfill entrypoint forwarded it.

### Calibrating the other three against source
- **DLQ lacks messageId/file context — does not reproduce at this head.** `internal/fileextract/dlq.go` `dlqRecord` already carries `MessageID`, `FileURL`, `FileExt`, `FileSize`, `Detail` (+ topic/partition/offset/key), with comments on replay-by-messageId. The context is present; no change needed here.
- **SIGTERM/cancellation classification — already handled.** `consumer.go` checks `ctx.Err()` at its loop boundaries (lines 96/101/117/192) and returns on cancellation rather than treating it as a processing failure. A dedicated graceful-vs-failure exit-code assertion would be a nice hardening, but it's not unhandled → at most a low 🟡.
- **`forbiddenSourceExcludes` exact-path vs wildcard — real but pre-existing 🟡.** `sourceExcludesContains` matches exact paths, so a `payload.file.*` wildcard exclude would evade the boot guard. This is the same defense-in-depth note already on record from the #46 arc, unchanged — non-blocking; worth a follow-up to match by prefix.

### Net
Verdict unchanged (Approve). One new non-blocking minor to fold in (backfill SSRF config parity), one pre-existing 🟡 (wildcard exclude), and two that are already covered at this head. None gate the code; merge remains blocked only on `check-sprint` (governance) + the human-review sign-off.

mochashanyao (approved):
[Octo-Q · automated review]

---

# Code Review — PR #47 (octo-search-indexer)

Reviewed at head `8c54213d979530536870ec67cb145a76d71a4a2b`. Scope: file content indexing pipeline (Kafka consumer + Tika extraction + OpenSearch partial update), backfill K8s Job, es-indexer scripted_upsert migration, SSRF defense, DLQ bounded retry.

## Summary

This PR introduces a complete file content indexing pipeline: a new `file-extractor` service consumes type=8 Kafka messages, downloads files from CDN (with SSRF double-gate defense), extracts text via Tika, and writes `payload.file.content` + `contentMeta` to OpenSearch via partial `_update`. A companion `file-content-backfill` K8s Job scrolls historical docs for one-shot backfill. The existing es-indexer is migrated from `index` to `update` + `scripted_upsert` with a Painless script that preserves file-extractor's fields across concurrent writers.

The architecture is well-structured: per-partition commit points prevent message loss, in-place bounded retry handles transient errors, DLQ bounded retry + spill-to-disk provides escape hatches, and tombstone markers prevent infinite backfill loops for permanent failures. Data-flow tracing across the full pipeline (Kafka → download → Tika → OS write → commit) confirms runtime data correctly reaches every consumption point.

## Verification

Static analysis only at head `8c54213d`; build and tests not executed in this environment.

- ✅ **Scripted upsert correctness** — `preservePainless` at `internal/esindex/writer.go:171` saves `content`/`contentMeta` from `ctx._source` before full-replace, then restores. Test `TestPreservePainless_ContainsAllPreservedPaths` asserts the script references all `preservedFilePaths`. The script handles `ctx.op == 'create'` (upsert path) separately with direct assignment.
- ✅ **Per-partition commit safety** — `partitionCommitPoints` at `internal/fileextract/decision.go` computes the longest consecutive-OK prefix per partition before committing. Transient messages break the prefix, preventing offset advancement past unprocessed messages.
- ✅ **SSRF defense completeness** — `validateURL` (scheme+host allowlist) → `ssrfRestrictedDialer` (resolve IPs, reject blocked) → `ssrfCheckRedirect` (re-validate on redirect). All three layers verified at `internal/fileextract/ssrf.go`. Blocked IP ranges cover RFC1918, loopback, link-local, CGNAT, IPv6 ULA, and cloud metadata (169.254.169.254).
- ✅ **Tombstone/scroll query alignment** — `tombstoneStatus` constant `"unextractable"` at `internal/fileextract/oswriter.go:101` matches `tombstoneStatusValue` at `internal/filebackfill/source.go:32`. Scroll query `must_not term contentMeta.status=unextractable` correctly excludes tombstoned docs.
- ✅ **Snowflake ID precision** — Both `parseHits` (backfill source) and `extractContentTypeFile` (payload parser) use `json.Decoder` + `UseNumber()` to avoid float64 truncation of 64-bit snowflake IDs.
- ✅ **UTF-8 boundary safety** — `truncateContent` at `internal/fileextract/tika.go:186` walks backward from the byte limit to find a valid UTF-8 boundary, preventing corrupted multi-byte characters.
- ✅ **DLQ idempotency** — DLQ writes happen before offset commit in `processBatch`. Crash-recovery may produce duplicate DLQ records, but they are harmless (same content, no side effects on downstream state).
- ✅ **`errDocNotYet` bounded retry** — Classified as `outcomeTransient` at `internal/fileextract/consumer.go:290`, retried up to `MaxRetriesPerMessage` (default 10), then forced DLQ with `ReasonRetryExhausted`. No infinite partition blocking.
- ✅ **`isPermanentStatus` 409 reclassification** — `internal/esindex/writer.go` now treats 409 as transient (was permanent). Correct: `retry_on_conflict=3` already retries server-side; remaining 409 may clear on consumer-level retry.

## Findings

No P0/P1 issues. Three P2 items and two nits below.

### P2 — New services lack OpenSearch TLS skip-verify transport (`internal/fileextract/oswriter.go:52`, `internal/filebackfill/source.go:58`)

The existing es-indexer consumer (`internal/consumer/service.go:77-79`), reconcile (`cmd/reconcile/main.go:99-100`), and old backfill (`cmd/backfill/main.go:112`) all support `ES_TLS_INSECURE_SKIP_VERIFY` via `esindex.InsecureSkipVerifyTransport()`. The new `oswriter` and `ossScrollSource` create `opensearch.Config` without a `Transport` field. In deployments using self-signed HTTPS certificates (clearly supported by the existing codebase), `assertLiveMapping` at startup will fail with a TLS certificate error, making the entire file-extractor service unable to start. The `ServiceConfig` struct at `internal/fileextract/config.go` has no `Transport` or TLS-related field to wire this through.

Diff-scope: new — PR adds services that lack an existing codebase capability. Fix: add `Transport http.RoundTripper` to `ServiceConfig`, populate from env in `cmd/file-extractor/main.go` and `cmd/file-content-backfill/main.go`, and pass through to both `opensearch.Config` constructors.

### P2 — Backfill exit-1 on any single OS transient is oversensitive at scale (`cmd/file-content-backfill/main.go`, `internal/filebackfill/runner.go:169-172`)

The exit code logic triggers exit 1 when `stats.OSTransient > 0`. For a 123K-doc backfill running ~40 minutes at 50 RPS, occasional 429s from OpenSearch rate limiting are expected. A single 429 during the entire run triggers exit 1, requiring operator investigation and a full re-run. Combined with `processOne` not retrying transient errors per-doc (unlike the Kafka consumer's in-place bounded retry), each transient error directly increments the counter with no mitigation. Consider either adding per-doc retry in `processOne` (2-3 attempts with short backoff) or changing the exit threshold to `OSTransient/Scanned > threshold` (e.g., 1%).

Diff-scope: new.

### P2 — `extract_timeout` and `download_failed` excluded from tombstones creates persistent backfill retry loop (`internal/fileextract/extractor.go`, `internal/fileextract/dlq.go`)

The `tombstoneReasons` whitelist deliberately excludes `download_failed` and `extract_timeout` to allow transient recovery. However, for files that are *permanently* undownloadable (e.g., CDN returns 404 consistently, file deleted at source), every backfill run will attempt extraction, fail, send to DLQ (logged), and skip — without writing a tombstone. The scroll query will re-select these docs on every subsequent backfill run. For 123K historical docs, even a small percentage of permanently-missing files produces repeated DLQ log noise across runs. Consider adding a counter-based tombstone mechanism (e.g., after N failed backfill attempts for the same doc, write a tombstone with reason `"repeated_download_failed"`).

Diff-scope: new. Not a merge blocker — flagging for operational awareness.

### Nit — `classifyOSErr` nil-response path wraps error generically (`internal/fileextract/oswriter.go:178`)

When `resp == nil && err != nil` (e.g., connection refused, DNS failure), the fallthrough wraps with `%w: %v` using `errOSTransient`. The original error is preserved but lacks the target host/URL context that would make the log actionable for on-call debugging. Consider including `cfg.ESAddresses` or the request index in the error message.

### Nit — `SSRFAllowLoopback` hardcoded false (`cmd/file-extractor/main.go`)

This is intentionally secure for production. For dev/test environments where the CDN runs on localhost, there's no override. Documenting the constraint or adding an env-configurable flag would ease local development.

## Things I checked that are fine

- **Painless script handles all edge cases**: `ctx._source.payload.file` as null, non-Map, or missing — script uses `instanceof Map` guards before accessing nested fields.
- **Rate limiter burst floor**: `newRateLimiter` at `internal/filebackfill/ratelimit.go:34` floors burst to 1, preventing deadlock when `docsPerSec < 1`.
- **`formatDuration` floors <1s scroll TTL to "1s"**: prevents invalid zero-duration scroll context at `internal/filebackfill/source.go`.
- **Content size cutoff**: `download.go` uses `io.LimitReader(maxSize+1)` + Content-Length pre-check, preventing unbounded downloads.
- **`doc_as_upsert: false` on tombstone writes**: prevents creating orphan docs when the main doc doesn't exist yet. `errDocNotYet` is handled gracefully (log + continue).
- **`encodedSingleDocSize` accounts for ~2x scripted_upsert inflation**: bulk batch splitting uses realistic size estimates, preventing oversized bulk requests.
- **Kafka manual commit with `CommitInterval=0`**: `FetchMessage` + `CommitMessages` gives precise offset control; no auto-commit risk.
- **DLQ handler spill-to-disk escape**: when DLQ writes fail and `SpillDir` is configured, records spill to JSON files on disk instead of blocking the consumer indefinitely.

## Verdict: APPROVED

No P1 blockers. The P2 items (TLS transport parity, backfill exit sensitivity, persistent tombstone gap) are non-blocking operational concerns that can be addressed in follow-up work. The core pipeline — data flow from Kafka through CDN download, Tika extraction, and OpenSearch partial update — is correctly implemented with proper error classification, bounded retry, and concurrent-writer safety via scripted_upsert.

yujiawei (approved):
# Code Review — PR #47 (octo-search-indexer)

Reviewed at head `8c54213d979530536870ec67cb145a76d71a4a2b` against `origin/main` (merge-base `7642bbd`). Build (`go build ./...`) and the full test suite for the three touched packages (`internal/fileextract`, `internal/filebackfill`, `internal/esindex`) pass locally.

This is a large, well-structured change (file-extractor consumer + backfill Job + es-indexer switch to `_bulk update`/`scripted_upsert`). The design is sound and the previously-flagged blocking issues appear resolved. I found **no P0/P1 blockers**; the items below are P2 suggestions and nits.

## Overall verdict: **Approve**

The core mechanisms are correct and I verified the load-bearing ones directly:

- **SSRF defense** (`internal/fileextract/ssrf.go`): dual gate (scheme/host allowlist pre-check + dial-time IP resolution blocking private/loopback/link-local/CGNAT/ULA/metadata) plus `CheckRedirect` re-validation. `isBlockedIP` correctly extends `net.IP.IsPrivate` to cover CGNAT `100.64/10` and IPv6 ULA `fc00::/7`. Solid.
- **Writer preserve semantics** (`internal/esindex/writer.go`): `scripted_upsert` + `preservePainless` correctly preserves `payload.file.content`/`contentMeta` on es-indexer redelivery; `retry_on_conflict=3` matches the sibling file-extractor writer; 409/429 are classified transient. `forbiddenSourceExcludes` + the startup `AssertLiveMappingCompatible` gate close the `_source.excludes` deployment-order trap. The byte-based sub-batching + virtual-child (`Derivatives`) handling is careful and tested.
- **Offset safety** (`internal/fileextract/consumer.go`, `decision.go`): per-partition contiguous-prefix commit is correct and avoids the cross-partition skip-commit data-loss class.
- **Error classification**: verified against `opensearch-go` v3.1.0 — `opensearchapi.(*Client).do` returns a **non-nil error on any non-2xx** (`resp.IsError()`), so `UpdateContent`/`WriteTombstone`/`Search` all route 4xx/5xx into `classifyOSErr`/the error path rather than silently succeeding. (I mention this because a status-only reading of the `resp != nil && err == nil → success` lines can look unsafe; it is safe given the client contract.)

## P2 — suggestions (non-blocking)

1. **`internal/fileextract/download.go:73`** — `http.Client.Timeout` is set straight from `cfg.DownloadTimeout` with no in-package default, unlike `maxSize`/`retries`/`backoff` just above. Production always sets it (30s), so this is defense-in-depth: a caller that constructs `ServiceConfig` without it gets a client with no overall timeout. Consider defaulting to 30s when `<= 0`.

2. **`cmd/file-content-backfill/main.go:134,147`** — `envInt`/`envFloat` use `fmt.Sscanf(v,"%d",…)` / `"%f"`, which accept partial matches: `BACKFILL_MAX_CONTENT_BYTES=256k` parses to `256` (bytes) with no error, and `0.1x` → `0.1`. This is a silent-misconfig footgun for a one-shot backfill. Prefer `strconv.Atoi` / `strconv.ParseFloat(v,64)` with full-string validation, consistent with `cmd/file-extractor/main.go` which already uses `strconv.Atoi`.

3. **`cmd/file-extractor/main.go` `loadConfig`** — with `FILE_EXTRACTOR_ENABLED=true` but a typo/omission in `KAFKA_BROKERS` or `ES_ADDRESSES`, `enabled` collapses to `false` and the pod idles silently (`Running` but processing nothing). Since the operator explicitly opted in, consider failing loud (non-zero exit / `log.Fatal`) when `ENABLED=true` yet mandatory deps are missing, rather than treating it as "not configured".

4. **`cmd/file-extractor/main.go:44`** — `if err != nil && ctx.Err() == nil` will swallow a genuine `run` error if a SIGTERM cancels the context in the same window, exiting 0 and hiding the failure from orchestration. Consider gating on the specific cancellation errors (`errors.Is(err, context.Canceled)/DeadlineExceeded`) instead of any `ctx.Err()`.

5. **`internal/filebackfill/source.go` `Close`** — the `DELETE /_search/scroll` response from `s.client.Client.Perform` is discarded with `_` and its `Body` is never closed (contrast `continueScroll`, which `defer`s the close). Minor connection leak, once per Job; worth closing for symmetry.

## Nits / notes

- **`internal/esindex/writer.go` `preservePainless`** — the preserve is unconditional for any existing doc. This is correct for the domain (snowflake `messageId` is never reused and a message's type is immutable, so a non-file doc never has saved file content and the branch is a no-op). Flagging only as a latent assumption: if message-id reuse or in-place type changes were ever introduced, stale `content` could survive onto a non-file doc.
- **Docs in a public repo** — `docs/file-content-indexing-{fix-plan,code-review,rereview}.md` embed the internal round-based review history (e.g. "Round-3 Blocker B", named reviewer initials). These are useful design records, but if this repo is public you may want to decide deliberately whether the internal review process/attribution belongs here vs. the architecture docs (`feasibility`/`implementation`/`tool-comparison`), which are clean.

None of the above blocks merge. Nice work consolidating the prior round of feedback.

yujiawei:
Follow-up to my approval — one additional non-blocking P2 worth capturing before this feature is enabled in production.

**P2 — new OpenSearch clients lack TLS skip-verify transport parity**

The existing OS-facing tools (`internal/consumer/service.go:77-79`, `cmd/reconcile/main.go:99-100`, `cmd/backfill/main.go:262`) all support self-signed HTTPS via `esindex.InsecureSkipVerifyTransport()`, gated on `ES_TLS_INSECURE_SKIP_VERIFY`. The three new OpenSearch construction sites do not:

- `internal/fileextract/oswriter.go:51` — `opensearch.Config` built with no `Transport`
- `internal/filebackfill/source.go:57` — same
- `cmd/file-extractor/main.go` `assertLiveMapping` — builds `esindex.Config` without `Transport`

Neither `fileextract.ServiceConfig` nor `filebackfill.Config` exposes a TLS/`Transport` field to wire it through. Since file-extractor targets the **same** OpenSearch cluster as es-indexer (and the backfill example here uses `https://10.10.148.x:9200`, i.e. internal-IP HTTPS), a deployment that relies on skip-verify would have es-indexer working while file-extractor fails its startup `AssertLiveMappingCompatible` on a TLS certificate error — a crashloop rather than a graceful degrade.

This does not block merge (the service is feature-gated off by `FILE_EXTRACTOR_ENABLED` and the es-indexer main path is unchanged), but it should be resolved before the feature is switched on in an environment that uses self-signed certs. Suggested fix: add `Transport http.RoundTripper` to both config structs, populate from `ES_TLS_INSECURE_SKIP_VERIFY` in `cmd/file-extractor` and `cmd/file-content-backfill`, and pass it into both `opensearch.Config` constructors (and the `esindex.Config` in `assertLiveMapping`).

Note the design docs already defer *Kafka* TLS/SASL to a separate PR; this OpenSearch-side gap is distinct and not covered by that note.

PR #50, opened 2026-07-08
feat(fileextract): split invalid_url from download_failed for SSRF pre-check permanent reject

aunknown:
## Bug

SSRF pre-check（`validateURL`: scheme/host allowlist）拒绝的 URL 被归到 `reason=download_failed`，而 `download_failed` 在 `tombstoneReasons` 白名单外（transient 保留 backfill retry safety net，见 `internal/fileextract/extractor.go` Round-4 review 注释）。结果 SSRF pre-check 拒的 URL 每次 backfill rerun 都会重复失败，无法收敛。典型触发场景：老历史 file 消息 URL 是已淘汰的 CDN/COS 域名，不在现役 allowlist 内。

## Fix

把 SSRF pre-check 拒绝拆到独立 permanent reason `invalid_url`：

- `internal/fileextract/download.go`: 新增 sentinel `errInvalidURL`
- `internal/fileextract/download.go` `Fetch()`: pre-check 失败改 wrap `errInvalidURL`（而非 `errDownloadFailed`）
- `internal/fileextract/extractor.go` `ExtractAndWrite()`: 在 `errDownloadFailed` 分支之前新增 `errors.Is(derr, errInvalidURL)` 分支返 `ReasonInvalidURL`
- `internal/fileextract/dlq.go`: 新增常量 `ReasonInvalidURL = "invalid_url"` + 表格文档
- `internal/fileextract/extractor.go` `tombstoneReasons`: 加 `ReasonInvalidURL: true`（allowlist 是硬 policy 边界，同 URL rerun 结果不变）

**未改**：dialer 层 SSRF 拒（`errSSRFPrivateIP` via `ssrfRestrictedDialer`）仍归 `download_failed` —— 走 tryFetch retry path，DNS/网络条件可能变化保守留 transient。仅 **前置 `validateURL` pre-check** 归 `invalid_url`。

## Test plan

- 新增 `TestExtractor_PermanentDLQ_WritesTombstone_InvalidURLHost` — URL host 不在 allowlist → reason=invalid_url + tombstone 写入 `status=unextractable` + `reason=invalid_url`
- 新增 `TestExtractor_PermanentDLQ_WritesTombstone_InvalidURLScheme` — URL scheme=ftp 非白名单 → 同上断言
- `TestTombstoneReasons_Whitelist` 契约锁 permanent list 加 `ReasonInvalidURL`
- `TestFetch_SSRFPrecheckBlocksMalformedURL` expect 从 `errDownloadFailed` 更新为 `errInvalidURL`
- `go vet ./...` + `go build ./...` 全绿
- `go test -race -timeout 120s ./internal/fileextract/... ./internal/filebackfill/...` 全绿

## 同仓 refs

- Round-4 `lml2468` review 关于 tombstoneReasons transient/permanent 拆分的注释：`internal/fileextract/extractor.go` `tombstoneReasons` var block 上方注释
- SSRF 双闸门实现：`internal/fileextract/ssrf.go` `validateURL` (闸门 1) + `ssrfRestrictedDialer` (闸门 2)
- DLQ reason 常量表：`internal/fileextract/dlq.go`

### DIFF
--- a/internal/fileextract/dlq.go
+++ b/internal/fileextract/dlq.go
@@ -11,6 +11,7 @@ package fileextract
 //	| oversize            | 文件 > MaxFileSize                            | ❌      |
 //	| blacklist_ext       | 扩展名在黑名单（不该走这里，是设计正常跳过） | ❌      |
 //	| download_failed     | HTTP GET CDN 5xx 或连接超时（重试耗尽）        | ✅ 3 次 |
+//	| invalid_url         | SSRF pre-check 拒 URL（scheme/host allowlist）| ❌      |
 //	| extract_timeout     | Tika 抽取 > ExtractTimeout                    | ❌      |
 //	| encrypted           | Tika 抛 EncryptedDocumentException             | ❌      |
 //	| empty_extract       | Tika 返回空串或空白                           | ❌      |
@@ -26,6 +27,11 @@ const (
 	ReasonEncrypted      = "encrypted"
 	ReasonEmptyExtract   = "empty_extract"
 	ReasonExtractError   = "extract_error"
+
+	// ReasonInvalidURL：SSRF pre-check 拒绝 URL（scheme/host 不在 allowlist）。
+	// 归 permanent DLQ + tombstone：allowlist 是硬 policy 边界，同 URL rerun 结果不变。
+	// 典型触发：老历史 file 消息 URL 是已淘汰的 CDN/COS 直连域名，不在现役 allowlist 内。
+	ReasonInvalidURL = "invalid_url"
 
 	// ReasonRetryExhausted：单条消息 in-place bounded retry N 次仍未成功
 	// (errDocNotYet / errOSTransient 长期未收敛) → 强制 DLQ 并 commit offset，避免
--- a/internal/fileextract/download.go
+++ b/internal/fileextract/download.go
@@ -31,6 +31,12 @@ var errOversize = errors.New("fileextract: file size exceeds MaxFileSize cutoff"
 // errDownloadFailed 是重试耗尽或 4xx 非 429 permanent 失败（触发 DLQ reason=download_failed）。
 var errDownloadFailed = errors.New("fileextract: download exhausted retries or 4xx permanent")
 
+// errInvalidURL 是 SSRF pre-check（validateURL）拒绝 URL 的 sentinel（触发 DLQ reason=invalid_url）。
+// 与 errDownloadFailed 分开：pre-check 拒是 policy 硬约束（scheme/host allowlist 不变，同 URL
+// 无论重试或 backfill rerun 结果不变），归 permanent tombstone；errDownloadFailed 保留给
+// CDN 5xx / 4xx expired 等 transient/URL-specific 故障（rerun 可能成功，不写 tombstone）。
+var errInvalidURL = errors.New("fileextract: url rejected by SSRF pre-check")
+
 // downloadClient 从 URL 拉文件 bytes。
 type downloadClient struct {
 	hc             *http.Client
@@ -94,8 +100,10 @@ func (c ServiceConfig) RetryBackoffBase() time.Duration {
 // URL 不变重试无意义）。
 func (d *downloadClient) Fetch(ctx context.Context, url string) ([]byte, string, error) {
 	if err := validateURL(url, d.allowedHosts, d.allowedSchemes); err != nil {
-		// SSRF pre-check 拒绝 → download_failed（不重试；host/scheme 不变重试无意义）
-		return nil, "", fmt.Errorf("%w: %v", errDownloadFailed, err)
+		// SSRF pre-check 拒绝 → invalid_url（不重试；host/scheme allowlist 是硬约束，同 URL rerun 无益）。
+		// 归 permanent tombstone reason（extractor.go tombstoneReasons），backfill scroll 会过滤，
+		// 避免老 URL（如已下线的历史 COS 直连）被 rerun 无限重试。
+		return nil, "", fmt.Errorf("%w: %v", errInvalidURL, err)
 	}
 	var lastErr error
 	for attempt := 0; attempt <= d.retries; attempt++ {
--- a/internal/fileextract/extractor.go
+++ b/internal/fileextract/extractor.go
@@ -59,6 +59,11 @@ func firstNonZero(a, b int64) int64 {
 // 跳过 → **永远不再重试**。commit message 声称的"permanent DLQ"列表也漏了 extract_timeout →
 // intent vs impl 不一致，本轮修 impl 匹配 intent（不改 intent）。
 //
+// v1.14：invalid_url 是从 download_failed 拆出的**永久** SSRF policy 类（validateURL pre-check
+// 拒），归 tombstone —— 与 download_failed 语义相反：URL host/scheme allowlist 是硬约束，
+// 同 URL rerun 结果不变（typical: 老历史消息带已淘汰的 CDN/COS 直连域名），需 tombstone
+// 让 backfill 跳过避免无限重复失败。
+//
 // 白名单只覆盖 extractor.ExtractAndWrite 内部返回的 permanent 类（不含 ParseError /
 // RetryExhausted / OSPermanent — 这些 reason 由 consumer.processBatch 直接 writeDLQ，不经
 // extractor.defer；ParseError 场景 doc 不存在于 OS 也不需要 tombstone）。
@@ -68,6 +73,7 @@ var tombstoneReasons = map[string]bool{
 	ReasonEncrypted:    true, // 加密 PDF：无密码无解
 	ReasonEmptyExtract: true, // Tika 抽出空/纯空白：文件真无文本
 	ReasonExtractError: true, // Tika 报 exception：内容/格式问题，重试无益
+	ReasonInvalidURL:   true, // SSRF pre-check 拒 URL：allowlist 硬约束，同 URL rerun 无益
 	// 明确排除：ReasonDownloadFailed（CDN 5xx / 网络抖动，rerun 可能成功）
 	// 明确排除：ReasonExtractTimeout（Tika 临时资源紧张 / 大文件超时，rerun 可能成功）
 }
@@ -130,6 +136,11 @@ func (e *Extractor) ExtractAndWrite(ctx context.Context, messageID string, fp *f
 	if derr != nil {
 		if errors.Is(derr, errOversize) {
 			return ReasonOversize, derr, nil
+		}
+		if errors.Is(derr, errInvalidURL) {
+			// SSRF pre-check 拒 → permanent (走 tombstone)。放在 errDownloadFailed 前判断，
+			// 否则 wrap 顺序变化时可能被更宽松的分支吃掉。
+			return ReasonInvalidURL, derr, nil
 		}
 		if errors.Is(derr, errDownloadFailed) {
 			return ReasonDownloadFailed, derr, nil
--- a/internal/fileextract/ssrf_test.go
+++ b/internal/fileextract/ssrf_test.go
@@ -163,7 +163,8 @@ func TestSSRFRestrictedDialer_LoopbackBypassWhenAllowed(t *testing.T) {
 }
 
 // TestFetch_SSRFPrecheckBlocksMalformedURL Fetch 入口 pre-check 拦 malformed URL，
-// 不重试（scheme/host 不变重试无意义）。
+// 不重试（scheme/host 不变重试无意义）。v1.14 起 pre-check 失败归 errInvalidURL
+// （从 errDownloadFailed 拆分，SSRF policy permanent 类，走 tombstone）。
 func TestFetch_SSRFPrecheckBlocksMalformedURL(t *testing.T) {
 	dc := newDownloadClient(ServiceConfig{
 		DownloadTimeout: 100 * time.Millisecond,
@@ -171,8 +172,8 @@ func TestFetch_SSRFPrecheckBlocksMalformedURL(t *testing.T) {
 		MaxFileSize:     1024,
 	})
 	_, _, err := dc.Fetch(context.Background(), "http://cdn.deepminer.com.cn/x") // http 被默认 scheme 拒
-	if !errors.Is(err, errDownloadFailed) {
-		t.Fatalf("expected errDownloadFailed wrapping SSRF, got %v", err)
+	if !errors.Is(err, errInvalidURL) {
+		t.Fatalf("expected errInvalidURL wrapping SSRF pre-check, got %v", err)
 	}
 }
 
--- a/internal/fileextract/tombstone_whitelist_test.go
+++ b/internal/fileextract/tombstone_whitelist_test.go
@@ -26,6 +26,7 @@ func TestTombstoneReasons_Whitelist(t *testing.T) {
 		ReasonEncrypted,
 		ReasonEmptyExtract,
 		ReasonExtractError,
+		ReasonInvalidURL, // v1.14：SSRF pre-check 拒 URL（scheme/host allowlist policy 硬约束）
 	}
 	for _, r := range permanent {
 		if !isTombstoneReason(r) {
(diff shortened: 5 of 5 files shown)


OctoBoooot (approved):
## Review: split invalid_url from download_failed for SSRF pre-check permanent reject (#50)

**Verdict:** Approve. Correct, well-scoped fix with sound security reasoning and a regression-locked contract. The change reclassifies SSRF pre-check rejections as a permanent `invalid_url` reason (tombstone, don't retry) while keeping dialer-layer private-IP rejects transient — and crucially does NOT weaken the SSRF gate itself. One precondition: `Test` is green; `code-review` check pending.

### Verified — security-critical properties
- **SSRF gate unweakened.** `download.go:102` still calls `validateURL(url, allowedHosts, allowedSchemes)` and returns on failure — only the error *wrapping* changed (`errInvalidURL` vs `errDownloadFailed`). A rejected URL is still never fetched; this only changes the DLQ reason, not the reject decision.
- **The transient/permanent split is the right distinction.** Pre-check allowlist rejection (`validateURL`: scheme/host) is a deterministic policy boundary — same URL fails every rerun → permanent `invalid_url` tombstone is correct (fixes the reported non-convergence where retired-CDN-domain history messages retried forever). Dialer-layer `errSSRFPrivateIP` stays `download_failed`/transient — DNS resolution can change, so keeping it retryable is the conservative, correct call. The PR draws this line explicitly and correctly.
- **`errors.Is` ordering correct.** `extractor.go:140` checks `errInvalidURL` before `errDownloadFailed`; the two are distinct sentinels (neither wraps the other) so ordering is independent here, but the defensive order + comment is right.
- **No fail-open.** The change is pure reason-reclassification of an already-rejected URL; no new fetch path.

### Tests — lock the exact contract
- `ssrf_test.go` retains full gate coverage: https-allowed, http-scheme reject, unknown-host reject, **metadata-IP-as-host reject**, malformed, empty-host, custom allowlist; dialer `DirectPrivateIPBlocked` → `errSSRFPrivateIP`.
- `tombstone_whitelist_test.go` is a **contract-lock**: asserts `ReasonInvalidURL` MUST tombstone (permanent) AND `ReasonDownloadFailed` MUST NOT (transient — preserves backfill retry), plus unknown-reason → conservative no-tombstone. This directly guards the invariant the fix depends on.

### Precondition (gate, not a finding)
`Test`/Build/Vet/Lint/CodeQL/osv/secret-scan all green; `code-review` check pending. Verdict is APPROVE; `check-sprint` red is governance (non-code), does not bar it.

### Praise
- Splitting on the *source of the rejection* (deterministic pre-check vs DNS-dependent dial-time) rather than lumping all SSRF rejects together is the correct, security-aware call — it makes permanent what is genuinely permanent without making a DNS-transient condition permanently un-retryable. And locking it with a whitelist-contract test (invalid_url tombstones, download_failed doesn't) is exactly how to keep a future refactor from silently flipping a transient reason to permanent (or vice versa).

Jerry-Xin (approved):
Summary: Project relevance gate passed; this PR is in scope for `octo-search-indexer` and cleanly separates SSRF pre-check policy rejects from transient download failures.

💬 Non-blocking
- 🔵 Suggestion: [internal/fileextract/download.go:99](/Users/mlamp/.openclaw/tmp/pr-review-5dfun1aa/repo/internal/fileextract/download.go:99) still documents pre-check failures as returning `errDownloadFailed`; update it to `errInvalidURL` to avoid misleading future maintainers.
- 🔵 Suggestion: [internal/fileextract/dlq.go:50](/Users/mlamp/.openclaw/tmp/pr-review-5dfun1aa/repo/internal/fileextract/dlq.go:50) still says “8” reason enums, and [internal/fileextract/dlq.go:58](/Users/mlamp/.openclaw/tmp/pr-review-5dfun1aa/repo/internal/fileextract/dlq.go:58) does not mention `invalid_url` as a reason that carries `FileURL`. These are documentation drift only.
- 🔵 Suggestion: [internal/fileextract/ssrf_test.go:165](/Users/mlamp/.openclaw/tmp/pr-review-5dfun1aa/repo/internal/fileextract/ssrf_test.go:165) names the case “MalformedURL”, but the input is a valid URL with a disallowed `http` scheme. Consider renaming the test or adding a truly malformed URL case.

✅ Highlights
- `errInvalidURL` is scoped to the pre-fetch `validateURL` path and does not change dialer-layer SSRF failures, matching the PR intent.
- `ExtractAndWrite` checks `errInvalidURL` before `errDownloadFailed`, preserving the new reason classification.
- `ReasonInvalidURL` is included in `tombstoneReasons`, so backfill can converge on policy-rejected historical URLs.
- Added tests cover host allowlist rejection, scheme rejection, and the tombstone whitelist contract.

Verification run:
- `go test ./internal/fileextract/... ./internal/filebackfill/...`
- `go vet ./...`
- `go build ./...`

mochashanyao (approved):
[Octo-Q · automated review]

---

# Code Review — PR #50 (octo-search-indexer)

Reviewed at head `df1307e34cdaf3813451c0ca4ea319ddc8769128`. Scope: SSRF pre-check sentinel split (`errInvalidURL` vs `errDownloadFailed`) with tombstone routing, ES TLS-insecure-skip-verify propagation across file-extractor / file-content-backfill, and Dockerfile binary additions.

## Summary

Solid, well-scoped PR that cleanly splits the SSRF pre-check rejection into a dedicated `errInvalidURL` sentinel so the DLQ layer can classify it distinctly from transient `errDownloadFailed`, and the new tombstone routing (`ReasonInvalidURL → permanent → writeTombstone`) correctly prevents backfill re-processing of host/scheme-rejected URLs. TLS skip-verify is propagated consistently across all four OpenSearch client construction sites using the shared `esindex.InsecureSkipVerifyTransport()`. No correctness, security, or build concerns.

## Verification

Static analysis only at head `df1307e3`; build and tests not executed in this environment.

- ✅ **End-to-end data flow** — `Fetch()` at `internal/fileextract/download.go:103` returns `errInvalidURL` wrapped; `ExtractAndWrite()` at `internal/fileextract/extractor.go:143` matches `errInvalidURL` **before** `errDownloadFailed` (correct ordering — prevents wrap confusion); returns `ReasonInvalidURL`; `isTombstoneReason=true` at `:76` triggers `WriteTombstone`; `attemptOne()` in `consumer.go` sees non-empty `dlqReason` → `writeDLQ` → `outcomeDLQ` → partition prefix advances.
- ✅ **Backfill skip** — `internal/filebackfill/source.go` scroll query uses `must_not term contentMeta.status=unextractable`, so tombstoned docs are correctly skipped on re-run.
- ✅ **TLS transport parity (C1)** — `ESTLSInsecureSkipVerify` flows through `internal/fileextract/config.go` → `internal/fileextract/oswriter.go` (newOSWriter), `internal/filebackfill/config.go` → `internal/filebackfill/source.go` (newOSScrollSource) + `ToExtractorConfig`, `cmd/file-extractor/main.go` (assertLiveMapping), and `internal/esindex` writer — all four OS client construction sites consistently inject `esindex.InsecureSkipVerifyTransport()` when the flag is true. No drift across symmetric sites.
- ✅ **Tests updated** — `ssrf_test.go` updated to expect `errInvalidURL`; `tombstone_whitelist_test.go` adds two new permanent-DLQ tombstone tests (host + scheme rejection); new `oswriter_tls_test.go` and `source_tls_test.go` cover both skip/strict branches.
- ✅ **Dockerfile** — adds build + COPY for `file-extractor` and `file-content-backfill`, consistent with existing binary pattern.

## Findings

No P0/P1 issues; two non-blocking items below.

### P2 — DLQ-constants guard test drift (`internal/fileextract/dlq.go:34`)

`ReasonInvalidURL` is added here, but `TestDLQReason_ConstantsExist` in `dlq_test.go` still hardcodes the v1.12 set of 8 reasons. The actual enum is now 11 (`+ReasonRetryExhausted`, `+ReasonOSPermanent`, `+ReasonInvalidURL`). If someone accidentally deletes any of the three new constants, the guard test will not catch it because the length-compare only trips when the test's own `want`/`got` slices disagree with each other — not with reality. Add the three missing constants to both slices to restore the test's stated purpose ("v1.12 §6.1 表格里的 8 种 reason 常量必须存在"). Not blocking — the runtime behavior is correct; only the guard is stale.

### Nit — Stale doc comments on `Fetch` (`internal/fileextract/download.go:103`)

The function-level comment on `Fetch` at line 97 still says the SSRF pre-check returns `errDownloadFailed`, but the code now returns `errInvalidURL`. Update to match. The package-level error-classification comment at `:10` also omits the `errInvalidURL` path — worth adding a line for completeness.

## Blind-spot checklist (C1–C6)

- **C1 (symmetric-path parity)** — CLEAR. TLS skip-verify propagation verified symmetric across all four OS client sites (oswriter, source, assertLiveMapping, esindex.Writer). Sentinel split has symmetric wrap/unwrap in `Fetch` and `ExtractAndWrite`.
- **C2 (control-flow ordering / nesting reuse)** — CLEAR. `errors.Is` chain in `extractor.go:143` orders `errInvalidURL` before `errDownloadFailed`; since `errInvalidURL` is never wrapped as `errDownloadFailed`, the ordering is defensive-correct. `attemptOne` in `consumer.go` is reused by both the retry state machine and the legacy `processOne` wrapper; permanent `dlqReason != ""` path is terminal in both.
- **C3 (authz boundary)** — N/A. No authz changes; TLS skip-verify is a client-side transport option scoped to self-signed OS clusters, gated behind explicit config flag / env var.
- **C4 (authz lifecycle / container-member cascade)** — N/A. No authz or container hierarchy touched.
- **C5 (build ≠ runtime correctness)** — CLEAR. Dockerfile adds binaries following the existing pattern; TLS transport injection is exercised by new `*_tls_test.go` tests covering both skip and strict branches against real TLS handshakes, not just build success.
- **C6 (governance / policy docs)** — N/A. No governance or policy docs changed.

## Data-flow trace

Consumed datum: `dlqReason` returned from `ExtractAndWrite` at `consumer.go:attemptOne`. Upstream: `extractor.go:143` returns `ReasonInvalidURL` when `errors.Is(derr, errInvalidURL)`; `derr` originates from `Fetch()` at `download.go:103` which wraps `errInvalidURL` on SSRF pre-check rejection. Runtime flow verified: SSRF reject → `errInvalidURL` wraps → `ExtractAndWrite` short-circuits to `ReasonInvalidURL` → `isTombstoneReason=true` → tombstone written → `attemptOne` routes via `writeDLQ` → `outcomeDLQ` → partition prefix commits. Backfill path: tombstoned doc carries `contentMeta.status=unextractable`; scroll query's `must_not term` skips it. No short-circuited or empty-data trap on this path.

## Cross-round Blocker Recheck

N/A — first review pass on this PR.

## Verdict: APPROVED

Only a P2 test-drift item and a nit-level stale comment — neither affects production correctness or the data-flow invariants this PR establishes. The SSRF sentinel split, tombstone routing, and TLS transport propagation are implemented cleanly and consistently across all symmetric sites.

[Octo-Q] verdict: APPROVE — No P0/P1 findings. The two non-blocking items (stale DLQ-constants guard test; stale doc comment) do not make any working path unavailable, produce wrong data, or regress prior behavior.

yujiawei (approved):
# Code Review — PR #50 (octo-search-indexer)

Reviewed at head SHA `df1307e34cdaf3813451c0ca4ea319ddc8769128`. Base `main`. Diff: 5 files, +135/-5.

Independent review plus two cross-model advisory passes (line-level + long-context). `go build ./...`, `go vet ./internal/fileextract/...`, and `go test -race ./internal/fileextract/...` all pass locally; the four new/updated tests pass.

## 1. Spec compliance

**Spec: ✅**

The change matches its stated intent exactly:

- **Not under-built.** All 5 declared edits are present: `errInvalidURL` sentinel (`download.go:38`); `Fetch()` pre-check now wraps `errInvalidURL` (`download.go:106`); `ExtractAndWrite` adds an `errors.Is(derr, errInvalidURL)` branch *before* the `errDownloadFailed` branch (`extractor.go:140-144`); `ReasonInvalidURL = "invalid_url"` constant + table doc (`dlq.go:34`, `dlq.go:14`); `tombstoneReasons[ReasonInvalidURL] = true` (`extractor.go:76`). Tests match the plan: two new tombstone tests (host + scheme reject), the whitelist contract test updated, and the existing pre-check test flipped from `errDownloadFailed` to `errInvalidURL`.
- **Not over-built.** No unrelated flags, fields, or interfaces added. The dialer-layer SSRF reject (`errSSRFPrivateIP`) is deliberately left as `download_failed` per the PR description; verified it is untouched.
- **No deviation.** Branch ordering is correct: `errInvalidURL` is checked before `errDownloadFailed`, and since `Fetch` only ever wraps one sentinel per return path they are not actually nested, so ordering is defensive rather than load-bearing — the inline comment says as much.

## 2. Code quality

**Quality: Approved**

Correct, well-tested, and internally consistent. No P0/P1. The findings below are advisory (P2) — worth a follow-up, not a merge blocker.

### P2 — Backfill cannot recover an `invalid_url` tombstone if a host is removed then re-added

`invalid_url` writes a tombstone (`contentMeta.status=unextractable`), and the backfill scroll query excludes on `status` alone, ignoring `reason` (`internal/filebackfill/source.go:141-144`). The allowlist is treated as a permanent policy boundary, which is true for the documented case (retired CDN/COS domains that never come back). But it is *config*, not an immutable property of the URL: if an operator temporarily drops a host/scheme from the allowlist, those docs get tombstoned, and re-adding the host later will **not** make backfill re-pick them — they stay invisible until the tombstones are manually cleared.

This is a recovery gap, not data loss: the raw message is still in Kafka/MySQL and the DLQ record carries the `FileURL` (`consumer.go:294`), so recovery is possible, just manual. Suggested follow-up (any one): let backfill optionally include `reason=invalid_url`, or add a small tombstone-clear tool keyed by reason, or document the "removing a host from the allowlist tombstones its docs" operational caveat next to `defaultAllowedDownloadHosts`. Since the whole point of the PR is convergence, making the un-converge path recoverable is the natural next step.

### P2 — Redirect-time SSRF rejects still classify as `download_failed`

`ssrfCheckRedirect` re-runs `validateURL` and returns a wrapped SSRF error (`ssrf.go:164-174`), but that error surfaces through `http.Client.Do` inside `tryFetch`, which treats all `Do` errors as transient (`download.go:149-155`) → retried → eventually wrapped `errDownloadFailed`. So a scheme/host policy reject that happens on a redirect gets the transient classification the PR is trying to eliminate for the pre-check path.

This is pre-existing (the PR does not touch the redirect path) and explicitly in the PR's stated scope ("only the front `validateURL` pre-check → `invalid_url`"). It is also arguably defensible: a redirect target is server-controlled and network-dependent, closer to the dialer layer that is intentionally left transient. Flagging so the asymmetry is a deliberate documented choice rather than an oversight — a one-line comment on `ssrfCheckRedirect` noting "redirect-time policy rejects stay download_failed by design" would close the loop.

### P2 — The reason split is invisible in aggregate metrics

Counters only expose aggregate `dlqTotal` (`metrics.go`, `IncDLQ`); there is no per-reason counter. Given `invalid_url` is now permanent, an allowlist regression that mass-tombstones live docs would be hard to detect from metrics alone — you'd have to consume DLQ records or grep logs. A per-reason (or at least an `invalid_url`) counter would materially reduce that operational risk. Pre-existing gap, but the permanence of the new reason raises its value.

### Nits

- `download.go:106` uses `fmt.Errorf("%w: %v", errInvalidURL, err)`. The `%v` stringifies the underlying `validateURL` error, dropping its type — you can `errors.Is(err, errInvalidURL)` but not unwrap to distinguish `errSSRFScheme` vs `errSSRFHost`. On Go 1.20+ consider `%w: %w` to preserve both, so future tests/telemetry can tell scheme rejects from host rejects.
- `ssrf_test.go` `TestFetch_SSRFPrecheckBlocksMalformedURL` is misnamed: the input `http://cdn.deepminer.com.cn/x` is well-formed and rejected purely on the `http` scheme not being allowlisted. Rename to reflect "disallowed scheme" (or test a genuinely malformed URL) to avoid confusing a future reader.

## 3. Overall verdict

**APPROVED**

Spec ✅ and Quality Approved. The fix is correct, tested, and does exactly what it claims. The three P2s are follow-up hardening (recovery path, redirect symmetry, observability), none of which blocks merge.

## 4. Suggestions

The single highest-value follow-up is making the `invalid_url` tombstone recoverable — a reason-aware backfill toggle or a targeted clear tool — so that an allowlist config mistake or rollback is not effectively permanent. Pair it with a per-reason DLQ counter so such a regression is visible before it accumulates.

## 5. Extra observations

Coverage note (both advisory legs ran successfully; this is not a "absence = pass"): I verified the diff, `ExtractAndWrite` branch ordering, `Fetch` classification, `validateURL`, the redirect path, tombstone write semantics, the backfill scroll filter, the DLQ record construction, consumer DLQ routing, and metrics. `dlqRecord.FileURL` is populated for the DLQ path so operators can locate affected URLs. What I did not verify: external metric/dashboard registration for the new reason, and any downstream DLQ-consumer/replay tooling that may switch on `reason` string values (worth a grep in the ops repo since a new enum value was introduced).

yujiawei:
Follow-up to the approved review — a first-pass automated check surfaced two additional non-blocking items, both confirmed on head `df1307e3`. Verdict stays **APPROVED**; noting them so they can ride along in this PR or a quick follow-up.

**P2 — DLQ-reason guard test is stale (`internal/fileextract/dlq_test.go:10-19`).** `TestDLQReason_ConstantsExist` still asserts the old 8-reason set. The enum is now 11: it's missing `retry_exhausted`, `os_permanent`, and the `invalid_url` this PR adds. The test's stated purpose is "these reason constants must exist (guard against accidental deletion/rename)", but it can't guard `invalid_url` (or the other two) because both its `want` and `got` slices omit them — deleting `ReasonInvalidURL` would not fail this test. Runtime behavior is correct; only the guard is blind. Add the three constants to both slices to restore its purpose.

**Nit — stale doc comments in `download.go`.** The `Fetch` doc comment (line 99) still says the SSRF pre-check returns `errDownloadFailed`; the code now returns `errInvalidURL`. The package-level error-classification comment (lines 8-11) also omits the new `errInvalidURL` path. Worth a one-line update each for the next reader.

Neither affects a working path, and both are safe to defer.

